Issue #38, opened 2026-03-12
Feature: make `/do` more quota-aware and support automatic or guided model downgrades


Dev-iL:
First off: thank you for releasing this tool, my experience with so far has been very positive! I really like the workflow and the quality bar that manifest-dev is aiming for.

I'm using the plugin on **Claude Code Team** with **Opus 4.6** in **high effort** mode. My concern is that there may be some built-in token inefficiency during `/do`, especially in longer verify/fix loops or when multiple sub-processes are active. I’m raising this mainly as a workflow-efficiency feature request. I think `/do` could benefit from treating model budget, quota budget, and context budget as three separate resources to optimize.

### Context
What I noticed is that the implementation phase (`/do`) seems to consume quota very quickly.

In one case, I had **two sessions waiting for quota reset**. After they resumed, usage jumped from **0% to 8% immediately**, and then climbed to around **70% over the next ~15 minutes**.  The plugin is clearly making progress during `/do`, but it feels like it may be spending premium reasoning budget on steps that do not always require it. If there’s a way to make execution more quota-aware, either automatically or via guided checkpoints, I think that would materially improve the experience for heavy users on Team plans.

Below are some suggestions my LLM made:

```md
## Why I think this may be worth looking at

Anthropic's current Claude Code surfaces already expose several concepts that seem relevant here:

- Claude Code supports **model switching mid-session** via `/model`.
- It also supports an **`opusplan`** mode that uses **Opus for planning** and **Sonnet for execution**.
- Claude Code supports **effort levels** (`low`, `medium`, `high`), and Anthropic explicitly notes that lower effort can reduce tool calls and overall token spend.
- Claude Code also has **status line** and **analytics** features that can expose model / context / usage / spend information.

That makes me think manifest-dev could potentially become smarter about **when full-power Opus + high effort is actually necessary**, versus when a cheaper/faster mode would be enough.

## Suggestion

### 1) Quota-aware execution during `/do`
If the underlying Claude Code APIs / interfaces make this possible, it would be great if manifest-dev could:

- monitor quota / spend / context pressure during `/do`
- automatically switch to a less expensive model when appropriate
- reserve the strongest model for the parts that actually benefit from it

For example:

- use **Opus** for:
  - architecture decisions
  - risk trade-offs
  - difficult debugging
  - ambiguous failures
- use **Sonnet** or **Haiku** for:
  - mechanical edits
  - retries after a narrowly scoped verification failure
  - boilerplate generation
[...]

PR #39, opened 2026-03-12
Claude: add quota-aware execution policy to /do and /verify

Dev-iL:
closes: #38 

## Summary

  This changes the Claude-specific manifest-dev workflow to carry an explicit execution policy through `/do`, `/verify`, and compaction recovery.

  The main behavior changes are:

  - `/do` now records an execution policy contract instead of treating quota usage as implicit
  - `/verify` now routes verification according to that policy, including economy-stage fan-out behavior
  - repeated verification failures now have clearer escalation guidance
  - post-compaction recovery restores execution policy from the execution log, including the fallback path
  - docs now describe the policy-aware workflow

  ## Scope

  This PR is intentionally Claude-focused.

  The changes live in Claude skill docs, Claude recovery behavior, and Claude-facing docs. They do not yet generalize the same workflow contract across other coding agents.

  To extend this to other agents such as Codex, the same three invariants should be implemented in each agent's own workflow layer:

  1. execution must persist an explicit policy
  2. verification must route/fan out based on that policy
  3. recovery/compaction must restore the policy from durable execution state

  That should be done in each agent's native instruction, recovery, and test surface rather than by copying Claude-specific skill text verbatim.

  ## Files Changed

  - `claude-plugins/manifest-dev/skills/do/SKILL.md`
  - `claude-plugins/manifest-dev/skills/verify/SKILL.md`
  - `claude-plugins/manifest-dev/hooks/post_compact_hook.py`
  - `tests/hooks/test_post_compact_hook.py`
  - `tests/test_skill_policy_contracts.py`
  - `claude-plugins/manifest-dev/README.md`
  - `README.md`

  ## Verification

  Passed:
  - `tests/test_skill_policy_contracts.py`
  - `tests/hooks/test_pretool_verify_hook.py`

  Also passed in local full-suite validation:
  - `tests/hooks/test_do_stop_hook.py`
  - `tests/hooks/test_post_compact_hook.py`

  Known unrelated red in full `pytest` on Windows:
  - `tests/test_dist_install_uninstall.py`
  - `tests/test_dist_skill_references.py`

  Those failures are preexisting and come from the root dist installer/reference tests invoking `bash` with Windows absolute paths. They are not caused by this PR's quota-awareness changes.

  ## Notes

  This PR changes workflow contract and recovery semantics, not the installer/distribution path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,14 @@ Then use it:
 
 Two commands. `/define` interviews you and builds a manifest. `/do` executes it. That's the whole workflow.
 
+If you want to tune execution intensity for a run, `/do` also accepts an optional `--policy=` flag:
+
+```bash
+/do <manifest-path> --policy=balanced
+```
+
+Supported modes are `economy`, `balanced`, and `max-quality`. If you omit the flag, behavior stays backward-compatible with the existing default. Policy guidance can recommend a stronger model or review path, but it does not automatically switch models or control quotas. Telemetry-aware guidance is optional and additive.
+
 If you use zsh and want easy upgrade commands for the non-Claude distributions, add this to `~/.zshrc`:
 
 ```zsh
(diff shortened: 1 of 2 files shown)


doodledood:
Introduced a variant of this as a part of the new version now. Thanks

Closing

PR #83, opened 2026-04-13
fix(stop_do_hook): escape valve when transcript flush lags behind /do→/done

tekicode:
## Problem

`stop_do_hook` blocks indefinitely on a long, self-contained `/do → /verify → /done` chain because Claude Code's transcript `.jsonl` flushes at user-prompt boundaries, not on every assistant turn. With no user prompt between `/do` and `/done`, the model's `Skill(skill: "manifest-dev:done")` invocation stays in memory and never reaches the on-disk file the hook reads.

The existing escape valve (`count_consecutive_short_outputs >= 3`) doesn't help: it relies on the same stale transcript and equally fails to see the model's `.` retries.

I hit this myself running `/manifest-dev:do` against a 4-deliverable manifest with a `--mode thorough` `/verify` that took ~5 minutes. After both `/verify` cycles passed and `/done` ran, the Stop hook fired hundreds of times, each reading a transcript frozen 7 hours in the past from the original `/do` invocation. Both the `/done` Skill call AND every `.` short-output retry I sent were buffered in memory, invisible to `parse_do_flow` and `count_consecutive_short_outputs` alike. Only an external user message could break the loop by triggering a transcript flush.

## Fix

Add a transcript-content-independent escape valve: count hook invocations themselves, keyed on transcript mtime. When N consecutive blocks all see the same mtime, allow with a diagnostic message that names both plausible causes:

  (a) Model's `/verify` or `/done` completed but transcript hasn't flushed
  (b) Model is genuinely stuck and not calling `/done` or `/escalate`

Both are user-actionable; neither is improved by the next block.

- New `record_stop_block(session_id, transcript_path, state_dir)` in `hook_utils.py`. State at `/tmp/manifest-dev-stop-blocks-<session_id>.json`. Counter resets when transcript mtime advances (so legitimate slow verifiers that DO flush eventually aren't false-escaped).
- New branch in `stop_do_hook.py` after the existing short-output check. Default threshold 5, override via `MANIFEST_DEV_STOP_MAX_STALE_BLOCKS`.
- 11 new tests: 5 escape-valve scenarios + 6 `record_stop_block` unit tests. All 53 stop_do_hook tests pass.

## Why count blocks instead of elapsed time?

A 15-minute `change-intent-reviewer` agent run never triggers Stop (we're not trying to stop, we're working). Block count stays 0 throughout. When `/verify` → `/done` → assistant attempts to stop, the buffer may still be 15+ minutes behind, but the hook can count its own invocations independently. Time-threshold approaches false-trip on legitimately slow verifiers; block-counting doesn't.

## Notes

- Drive-by: added `from pathlib import Path` for a pre-existing F821 in the test file's `tmp_path: Path` annotation.
- Ruff clean on all touched files. Mypy not run (local environment issue with `pathspec.patterns.gitignore`).
- Pre-existing `tests/test_dist_skill_references.py` failures are unrelated to this change (confirmed on baseline).

### DIFF
--- a/claude-plugins/manifest-dev/hooks/hook_utils.py
+++ b/claude-plugins/manifest-dev/hooks/hook_utils.py
@@ -10,6 +10,8 @@ from __future__ import annotations
 import json
 import os
 import re
+import sys
+import tempfile
 from dataclasses import dataclass
 from typing import Any
 
@@ -307,7 +309,7 @@ def count_consecutive_short_outputs(transcript_path: str) -> int:
 def record_stop_block(
     session_id: str,
     transcript_path: str,
-    state_dir: str = "/tmp",
+    state_dir: str | None = None,
 ) -> int:
     """
     Track consecutive stop-hook blocks per session, keyed on transcript mtime.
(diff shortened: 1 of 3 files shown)


tekicode:
Closing — the author addressed this differently in 3370af0: broadened the idle classifier to remove the `text_len >= 100` threshold (which was the practical root cause for background-agent deadlocks) and rewrote the systemMessage strings. Simpler fix that covers the common case. Our mtime-based escape valve addressed a different edge (transcript flush lag) but isn't needed with the improved classifier.

— Scout 🐾

PR #139, opened 2026-05-15
Add argument-hint frontmatter to user-invocable skills

doodledood:
## Summary

Adds `argument-hint` to 15 SKILL.md frontmatters across four plugins so the `/` autocomplete UI surfaces expected positional and flag shape per the [Claude Code skills docs](https://code.claude.com/docs/en/skills#frontmatter-reference).

- Strict convention: `<required>` for required positionals, `[optional]` for everything else.
- Flag values shown as placeholders (`<level>`, `<name>`, `<ids>`, `<pr-url>`) rather than enums so hints stay readable in autocomplete; valid values are documented in each skill body and enforced by the skill's argument parser.
- `user-invocable: false` skills skipped (no autocomplete UI to render the hint).
- Patch version bump on each affected plugin (metadata-only change).
- `dist/{gemini,opencode,codex}` SKILL.md mirrors propagated; sync-meta advanced to track this sync.

### Hints by skill

| Skill | Hint |
|---|---|
| `do` | `<manifest-path> [--mode <level>] [--scope <ids>]` |
| `verify` | `<manifest-path> [--mode <level>] [--scope <ids>] [--deferred]` |
| `auto` | `[task] [--mode <level>] [--platform <name>] [--babysit <pr-url>]` |
| `define` | `[task] [--mode <level>] [--interview <style>] [--amend <path>] [--babysit <pr-url>] [--platform <name>] [--canvas]` |
| `figure-out` | `[topic] [--with-docs]` |
| `adr` | `<manifest-path> <output-dir> --session <transcript-path>` |
| `walk-pr` | `[pr-url-or-range] [--canvas]` |
| `prompt-engineering` | `<request>` |
| `handoff` | `[prior-handoff-path]` |
| experimental `auto` / `define` | `[task] [--babysit <pr-url>] [--platform <name>] [--canvas]` |
| experimental `do` / `verify` | `<manifest-path> [--scope <ids>]` |

## Test plan

- [ ] Type `/do `, `/auto `, `/define ` in Claude Code and confirm the hint renders in the autocomplete row
- [ ] YAML parses cleanly for all 15 SKILL.md (verified locally via `yaml.safe_load`)
- [ ] `ruff check claude-plugins/` passes
- [ ] No behavior changes — argument-hint is metadata only

https://claude.ai/code/session_01M6TPj5zNGZ6yo1jybaadBD

---
_Generated by [Claude Code](https://claude.ai/code/session_01M6TPj5zNGZ6yo1jybaadBD)_

### DIFF
--- a/claude-plugins/PLUGIN_TEMPLATE/.claude-plugin/plugin.json
+++ b/claude-plugins/PLUGIN_TEMPLATE/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "plugin-template",
   "description": "A template for creating new Claude Code plugins",
-  "version": "1.2.0",
+  "version": "1.2.1",
   "author": {
     "name": "Your Name",
     "email": "your.email@example.com"
--- a/claude-plugins/PLUGIN_TEMPLATE/skills/example/SKILL.md
+++ b/claude-plugins/PLUGIN_TEMPLATE/skills/example/SKILL.md
@@ -1,6 +1,7 @@
 ---
 name: example
 description: 'Analyzes the current project structure and tech stack. Use when asked to explore, understand, or summarize a project. Trigger terms: project overview, analyze codebase, what is this project.'
+argument-hint: '<input>'
 ---
 
 # Example Skill
--- a/claude-plugins/manifest-dev-experimental/.claude-plugin/plugin.json
+++ b/claude-plugins/manifest-dev-experimental/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "manifest-dev-experimental",
-  "version": "0.10.0",
+  "version": "0.10.1",
   "description": "Experimental parallel rework of manifest-dev's skills. Radically slim prompts that trust model capability — every word steers, not scaffolds. Drops multi-mode budgeting and prescriptive HOW in favor of WHAT/WHY. Lives alongside manifest-dev until promotion. Same skills (figure-out, define, do, verify, done, escalate, auto), same hooks, same agents.",
   "keywords": [
     "experimental",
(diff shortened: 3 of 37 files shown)


Dev-iL:
Isn't the `adr` hint a bit strange? Seems like (the literal) `--session` shouldn't be required/present or all of it wrapped in `[]`

```bash
# current
<manifest-path> <output-dir> --session <transcript-path>
# suggested options
<manifest-path> <output-dir> <transcript-path>
<manifest-path> <output-dir> [--session <transcript-path>]
```

Another thought I had - since the interview part is now largely in `figure-out`, perhaps `adr` should get the figure-out session path instead of the define one (which is no longer interactive)

doodledood:
You are right 

Done in https://github.com/doodledood/manifest-dev/pull/140

doodledood:
@Dev-iL Btw, did you try the new --with-docs of figure-out?

I've taken it from Matt Pocock grill me with docs skill currently seems nice 

It creates ADRs on demand during the figure out like you wanted it originally but it fits more naturally there rather in the define 

In general it seems to me at least and pivoting there as a whole to first dig around in a more Freeform in a figure out session and when you are really on it only then define which becomes thinner and embeds the decisions into a formal manifest that is executable 

I now always start every session with figure out

Dev-iL:
> @Dev-iL Btw, did you try the new --with-docs of figure-out?
> 
> I've taken it from Matt Pocock grill me with docs skill currently seems nice
> 
> It creates ADRs on demand during the figure out like you wanted it originally but it fits more naturally there rather in the define
> 
> In general it seems to me at least and pivoting there as a whole to first dig around in a more Freeform in a figure out session and when you are really on it only then define which becomes thinner and embeds the decisions into a formal manifest that is executable
> 
> I now always start every session with figure out

I did try `/figure-out`, but never with `--docs`. I semi followed the recent changes, and found it a bit surprising that define stopped asking me questions. So, does `--docs` only generate ADRs, or is it a prereq for `/define`?

Btw, is there an entry point that starts from a "skill-less" plan mode? Should I use a plan as a starting point for figure-out? Perhaps we should have a `--from-plan` kwarg to define/figure-out?

doodledood:
Now define should invoke figure out id it doesnt yet have enough context in the experimental set of skills which are probably going to replace the existing ones (simplified, less token consumption and little tweaks) 

--with-docs should both maintain a context file in the repo that has vocabulary and propose ADRs under docs/adrs by default

Probably should pair that in a repo with pointers to those directories in the CLAUDE.md so every session benefits regardless of building something or not @Dev-iL

PR #156, opened 2026-05-20
fix(figure-out-team): force /loop polling at 2-min default

doodledood:


Dev-iL:
Consider setting an exponential backoff interval instead of a constant one (upon detecting no new replies)

PR #172, opened 2026-06-04
figure-out: domain probing via its own probe task files

doodledood:
## Why

A `/figure-out` session investigating why the model doesn't naturally discuss **how a task will be verified** during understanding (a real fail point: verification was never raised, so the design lacked the seams to verify it, forcing post-hoc test-affordance rework).

The investigation found a larger structural gap: the 1.0.0 over-slim (`17043f9`) **evicted** domain probing fuel from `define/tasks/*` and declared probing to be figure-out's job — but figure-out was never given the probes or a mechanism to load them. Verification was one symptom of a missing probing layer.

## What changed

- **ADR** `docs/adr/20260604-figure-out-owns-domain-probing-via-mirrored-task-files.md` — figure-out owns domain probing via its own per-domain probe task files (mirroring define's taxonomy), each skill carrying its **own self-contained detection index** so figure-out runs standalone. Calibrated re-homing of proven non-natural probes, not a wholesale restore.
- **figure-out probe task files** — new `skills/figure-out/tasks/` with `CODING.md`, `FEATURE.md`, `BUG.md`, `REFACTOR.md`. Each file: a "suggestions, not an agenda" header, a question-shaped **Blind-spot probes** section (non-natural angles only — verification promoted to a first-class, design-shaping probe), and a **Forced trade-offs** section. Sourced from the pre-strip task-file content but filtered to angles the model skips by default.
- **figure-out `SKILL.md`** — consumption line: when the topic is a code change, load the matching probe file (puts angles in view), then press only what's load-bearing by judgment — awareness, not a script; general probing otherwise. The domain→file mapping is **inlined** here (no separate index file).
- **Index inlined per PR review** — the task-type detection index lives inline in each consuming `SKILL.md` rather than a separate `tasks/README.md` (which the agent always loaded anyway); both index READMEs removed. define's `SKILL.md` carries the full Domains table + composition rules + content-types.
- **Docs reconciled to the two-parallel-sets model** — `CLAUDE.md`, root + plugin `README.md`, and the `CONTEXT.md` `Task File` definition: figure-out's task files carry probes; define's carry Quality Gates + Defaults; each skill owns its own inline index.
- **Plumbing** — dist regenerated for OpenCode and Codex (CLI-specific substitutions preserved); plugin version bumped 2.2.0 → 2.3.0.

## Verification

Built via `/auto` (figure-out → define → do), then amended for the inline-index review. Every Acceptance Criterion and Global Invariant verified by a subagent: probe-file structure, non-natural-only content, verification-first-class, inline self-contained index (no define coupling, no leftover README), consumption framing, doc reconciliation, project gate (ruff/black/mypy green), version + README sync, and dist consistency across both distributions.

https://claude.ai/code/session_011iqfzZo1JLjfFvJ3wK1w6W


claude-plugins/manifest-dev/skills/figure-out/SKILL.md
```diff
@@ -7,6 +7,8 @@ user-invocable: true
 
 Press the topic relentlessly. Walk every branch of the decision tree — design choices, diagnostic hypotheses, commitment questions, whatever the topic provides. Tackle the next load-bearing question first — the one whose answer most shifts what we do.
 
+When the topic maps to a known domain, load its probe file from `tasks/` (the index in `tasks/README.md` maps indicators to files) to surface angles that are easy to under-weight — verification among them. Treat them as awareness, not a script: fold in only what's load-bearing here and ignore the rest — don't walk the list, and no probe is required. No domain match → probe generally, as you would anyway.
```
doodledood: This is redundant. Inline the index. The agent will always load it. Same for define and figure out

claude-plugins/manifest-dev/skills/figure-out/SKILL.md
```diff
@@ -7,6 +7,8 @@ user-invocable: true
 
 Press the topic relentlessly. Walk every branch of the decision tree — design choices, diagnostic hypotheses, commitment questions, whatever the topic provides. Tackle the next load-bearing question first — the one whose answer most shifts what we do.
 
+When the topic involves a code change, load the matching probe file from `tasks/` to surface angles that are easy to under-weight — verification among them: `CODING.md` (any code change) composed with `FEATURE.md` (new functionality, APIs), `BUG.md` (defects, regressions), or `REFACTOR.md` (restructuring, cleanup) when one applies. Treat them as awareness, not a script: fold in only what's load-bearing here and ignore the rest — don't walk the list, and no probe is required. No code change, or no file fits → probe generally, as you would anyway.
```
doodledood: Write as a table not a paragraph. So it's easier to edit

claude-plugins/manifest-dev/skills/figure-out/SKILL.md
doodledood: Done. Inlined the detection index into both `SKILL.md`s and deleted both `tasks/README.md` files — figure-out keeps its compact domain mapping in the consumption line; define gets the full Domains table + composition rules + content-types folded in. Dist + CLAUDE.md/CONTEXT.md wording reconciled to match.

---

claude-plugins/manifest-dev/skills/figure-out/SKILL.md
doodledood: Done — rendered the figure-out domain mapping as a table (same style as define's), framing and graceful-degradation lines unchanged.

---

PR #183, opened 2026-06-07
Zero-agent rework: review-code skill, Codex plugin-native, Pi command rename + two-package split

doodledood:
## What & why

Eliminates the Pi skill-leak at its root and consolidates the reviewer architecture into skills. The legacy Codex installer copied portable skills into the shared `~/.agents/skills/`, which Pi scans globally — so manifest-dev commands surfaced in every Pi session. Codex plugins install into the Pi-invisible plugin cache, so migrating Codex to plugin-native removes the leak. Along the way, **manifest-dev now ships zero agents** — every former agent is a skill — the Pi commands lose their `manifest-` prefix, and the Pi runtime splits into two packages mirroring the Claude/Codex plugin split.

## Changes

**1. Zero agents; `verify.agent` removed from the manifest schema**
- The 13 quality-dimension reviewer agents became one `review-code` skill — **one dimension per invocation**, each loading exactly one reference file (progressive disclosure).
- The former functional agents are now skills too, **verb-named**: `github-pr-lifecycle` → `check-pr`, `slack-poller` → `poll-slack`, `prompt-reviewer` → `review-prompt`; `criteria-checker` dropped (a general-purpose verifier following `verify.prompt` already does single-criterion checks). `code-review` → `review-code`.
- There is no `verify.agent` field anymore. Every gate is verified by a **general-purpose** subagent whose `verify.prompt` *activates* the relevant skill (e.g. *"Activate the manifest-dev:review-code skill with dimension=code-bugs…"*) — it must not spawn a nested agent. `/define` encodes this automatically.
- `prompt-engineering` skill now documents defaulting to a skill over an agent (cross-harness compatibility) and the verb/noun naming convention.

**2. Pi command rename → `/do`, `/auto`, `/babysit-pr`** (drop the `manifest-` prefix), across the extension, tests, `pi-cli.md`, CONTEXT.md, and READMEs.

**3. Codex plugin-native distribution (installer retired)**
- `dist/codex` is now two native plugins (`manifest-dev`, `manifest-dev-tools`) with `.codex-plugin/plugin.json` + bundled skills, registered by repo-root `.agents/plugins/marketplace.json` (`authentication: ON_INSTALL`). The `install.sh` / `install_helpers.py` / `config.toml` merge / `rules/` / `agents/*.toml` installer is gone. Install: `codex plugin marketplace add doodledood/manifest-dev` **then** `codex plugin add manifest-dev@manifest-dev` and `…-tools@manifest-dev`.

**4. Pi runtime split into two packages**
- `@doodledood/manifest-dev-pi` (core: `/do`, `/auto` + shared Harness-level Do runtime) and `@doodledood/manifest-dev-pi-tools` (`/babysit-pr`, depends on core via `file:../..`, reuses its wiring; a command-ownership filter prevents double-verification).
- The repo-root `pi.extensions` lists **both** extensions so one `pi install git:…@main` registers `/do`, `/auto`, and `/babysit-pr`.
- **Verifier flags have a single owner per process** via a `globalThis` marker (shared across Pi's fresh per-extension module instances): the first extension to load registers `--manifest-verifier-*` (one `pi --help` entry); the other reads parsed values via a `process.argv` fallback. Works in both the repo-root and standalone-tools load paths.

**5. Verifier-spawn resilience**
- Each `subagents.spawn` is retried once after yielding a tick (a spawn can transiently hit Pi's stale-session-context guard at the executor checkpoint while the session settles). If no verifier spawns at all, Harness verification reports **one** harness/runtime-orchestration BLOCKED (underlying spawn error + session-id diagnostics) instead of an identical BLOCKED on every gate that never ran.

**6. OpenCode** stays installer-based — only Codex migrates. `review-code` added; dimension reviewer agents dropped; the dist `define/tasks/CODING.md` correctly references `AGENTS.md`.

**7. Docs/versioning**: CONTEXT.md language, READMEs, `CHANGELOG.md` migration notes.

## Breaking changes
- Reviewers are no longer addressable by agent name — use the `review-code` skill dimension (`/define` encodes this automatically); `verify.agent` is removed.
- Codex installs via the plugin marketplace (two-step), not the retired installer.
- Pi commands dropped the `manifest-` prefix.

See `CHANGELOG.md` for migration detail.

## Versions
- `manifest-dev` plugin: 2.4.0 → 2.8.0
- `manifest-dev-tools` plugin: 0.19.0 → 0.23.0
- `@doodledood/manifest-dev-pi` (and new `@doodledood/manifest-dev-pi-tools`): 0.8.4

## Verification
- `ruff` / `black --check` / `mypy` — clean
- `pytest tests/` — 32 passed
- `node --test` Pi runtime — 39 passed (core + tools + flag-ownership + spawn resilience)

---
_Generated by [Claude Code](https://claude.ai/code/session_01VpoKL9Qz4h5M8qQUxFABeJ)_


packages/manifest-dev-pi-tools/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "manifest-dev-tools",
+    "agent-skills"
+  ],
+  "pi": {
+    "extensions": [
+      "./pi/extensions/manifest-dev-tools.ts"
+    ]
+  },
+  "dependencies": {
+    "@doodledood/manifest-dev-pi": "0.7.0"
```
aviramkofman: This dependency makes the documented Pi git install fail before Pi can load anything. `packages/manifest-dev-pi-tools/package.json:19` depends on `@doodledood/manifest-dev-pi@0.7.0`, but the package with that name is the private repo root and is not included by the `packages/*` workspace, so npm resolves it from the public registry and hits the E404 reproduced by `npm install --dry-run`. Make the core package a real workspace dependency, use a local/file dependency that resolves during git installs, or remove the workspace dependency from the repo-root install path.

package.json
```diff
@@ -1,14 +1,21 @@
...
     "pi-package",
     "manifest-dev",
     "agent-skills"
   ],
+  "exports": {
+    "./extension": "./pi/extensions/manifest-dev.ts",
+    "./runtime": "./pi/extensions/manifest-dev-runtime.ts"
+  },
   "pi": {
     "extensions": [
```
aviramkofman, line 20: The repo-root Pi package still only loads the core extension at `package.json:20`, so the documented `pi install git:github.com/doodledood/manifest-dev@main` never registers `/babysit-pr`. Pi loads resources from the installed package root's `pi` manifest; workspaces and dependencies are not additional resource roots, so the tools package extension is invisible unless it is listed here or installed separately. Add the tools extension to the root Pi manifest with a repo-valid import path, or change the install flow to install the tools package as its own Pi package.

pi/extensions/manifest-dev.ts
```diff
@@ -193,22 +205,22 @@ export async function startManifestDo(
...
-			? "Usage: /manifest-auto <task>"
-			: "Usage: /manifest-babysit-pr <github-pr-url>";
+		const usage = command === "auto"
+			? "Usage: /auto <task>"
+			: "Usage: /babysit-pr <github-pr-url>";
 		ctx.ui.notify(usage, "warning");
 		return;
 	}
-	if (command === "manifest-babysit-pr" && !isGithubPr(args)) {
+	if (command === "babysit-pr" && !isGithubPr(args)) {
```
aviramkofman: `/babysit-pr` validates the entire raw argument string as a GitHub PR URL at `pi/extensions/manifest-dev.ts:223`, so supported invocations like no args for PR inference, `--manifest <path>`, and `--ci <url>` are rejected before the babysit workflow can run. Parse the babysit flags before URL validation and pass the manifest/CI context through, or narrow the advertised Pi grammar to match this wrapper.

claude-plugins/manifest-dev/skills/define/SKILL.md
```diff
@@ -24,12 +24,23 @@ Encode the conversation's shared understanding as a Manifest at `~/.manifest-dev
...
 
 **Verifier prompt discipline.** Before writing `verify.prompt` fields, invoke the prompt-engineering skill if it is available; if not, apply its core discipline inline. Verifier prompts are prompts: state the verifier's goal, evidence to inspect, PASS/FAIL/BLOCKED contract, and non-obvious context. Do not run a separate prompt-engineering interview — /define owns the manifest interview.
 
+**Encoding gates — always general-purpose + a skill.** There is no `verify.agent` field. Every gate is verified by a **general-purpose** subagent driven by `verify.prompt`; when a gate needs specialized behavior, the prompt tells that agent to **activate a skill**. Code-quality gates activate the `review-code` skill with a dimension; other specialized checks activate their skill (`check-pr`, `review-prompt`). Pattern:
+
+```yaml
+verify:
+  prompt: |
+    Spawn a general-purpose review using the manifest-dev review-code skill with dimension=<dimension> against the change.
```
aviramkofman: `define/SKILL.md:38` puts “Spawn a general-purpose review” inside the `verify.prompt` that `/do` already hands to a general-purpose verifier. Generated gates will ask the verifier to create a nested agent instead of activating `review-code` itself, which bypasses the runtime-owned verifier contract and can lose PASS/FAIL/BLOCKED evidence. Change the templates to tell the current verifier to activate `manifest-dev:review-code` with the dimension, and make the same wording change for the `check-pr` examples.

dist/opencode/skills/define/tasks/CODING.md
```diff
@@ -4,36 +4,38 @@ Base guidance for all code-change tasks (features, bugs, refactors).
 
 ## Quality Gates
 
-AGENTS.md may specify project-specific preferences.
+CLAUDE.md may specify project-specific preferences.
```
aviramkofman: `dist/opencode/skills/define/tasks/CODING.md:7` now tells OpenCode to read `CLAUDE.md`, but the OpenCode distribution installs `AGENTS.md` as its context file and the previous generated dist used `AGENTS.md`. OpenCode users will miss project gates unless a repo happens to carry Claude docs too. Keep the Claude source as `CLAUDE.md`, but fix the OpenCode conversion so this generated task guidance, the adherence row, and the defaults refer to `AGENTS.md`.

README.md
```diff
@@ -19,18 +19,18 @@ Three skills, one idea: figure it out, write down what "done" means, let it buil
...
 /plugin install manifest-dev@manifest-dev-marketplace
 
-# OpenCode — skills, agents, commands, plugin
+# OpenCode — skills, commands, plugin
 curl -fsSL https://raw.githubusercontent.com/doodledood/manifest-dev/main/dist/opencode/install.sh | bash
 
-# Codex CLI — skills, TOML stubs, rules, config
-curl -fsSL https://raw.githubusercontent.com/doodledood/manifest-dev/main/dist/codex/install.sh | bash
+# Codex CLI — native plugins from the repo marketplace
+codex plugin marketplace add doodledood/manifest-dev
```
aviramkofman, line 26: `README.md:26` only adds the Codex marketplace, but Codex separates marketplace registration from plugin installation, so this quick start leaves both plugin bundles uninstalled. Add explicit install commands for both plugins after the marketplace add, for example `codex plugin add manifest-dev@manifest-dev` and `codex plugin add manifest-dev-tools@manifest-dev`, and update the uninstall instructions to remove installed plugins rather than only removing the marketplace source.

packages/manifest-dev-pi-tools/package.json
doodledood: Fixed in `e01d0c3`: changed the tools→core dependency to `"file:../.."` so git installs resolve the private repo-root package locally instead of hitting the public registry (the E404). Test updated to assert the file-path dep.

---

package.json
doodledood: Fixed in `e01d0c3`: the repo-root `pi.extensions` now lists **both** extensions (`./pi/extensions/manifest-dev.ts` and `./packages/manifest-dev-pi-tools/pi/extensions/manifest-dev-tools.ts`), so a single `pi install git:...@main` registers `/do`, `/auto`, and `/babysit-pr`. The command-ownership filter keeps the two runtimes from double-verifying each other's runs.

---

pi/extensions/manifest-dev.ts
doodledood: Fixed in `e01d0c3`: the babysit wrapper now extracts the PR URL token from the args (so `--ci` and `--manifest <path>` no longer fail validation), rejects only when no URL is present, and threads `--ci` (CI one-shot cadence) and `--manifest <path>` (use existing manifest, skip synthesis) through `buildManifestBabysitPrompt`.

---

claude-plugins/manifest-dev/skills/define/SKILL.md
doodledood: Good catch — fixed in `e01d0c3`. The templates now say *"Activate the manifest-dev:review-code skill with dimension=X and review the change"* (and *"Activate the manifest-dev check-pr skill"*), so the current general-purpose verifier activates the skill directly instead of spawning a nested agent. Updated `define/SKILL.md`, `tasks/CODING.md`, `tasks/PR_LIFECYCLE.md`, and the README examples, with a note in the encoding guidance explaining why; re-mirrored to all dists.

---

dist/opencode/skills/define/tasks/CODING.md
doodledood: Fixed in `e01d0c3`: the OpenCode dist's `define/tasks/CODING.md` now refers to `AGENTS.md` for project preferences, project gates, the adherence gate-row, and the defaults (the Claude source stays `CLAUDE.md`). Also clarified `opencode-cli.md`'s context-file rule that CODING.md is operational (not comparative) so future syncs remap it; the Codex dist already used `AGENTS.md` and is now consistent.

---

README.md
doodledood: Fixed in `e01d0c3`: the Quick Start (and `codex-cli.md` + `dist/codex/README.md`) now does `codex plugin marketplace add doodledood/manifest-dev` **then** `codex plugin add manifest-dev@manifest-dev` and `codex plugin add manifest-dev-tools@manifest-dev`. Uninstall updated to `codex plugin remove manifest-dev && codex plugin remove manifest-dev-tools && codex plugin marketplace remove manifest-dev`.

---

pi/extensions/manifest-dev.ts
```diff
@@ -333,18 +351,30 @@ Follow the manifest-dev chain autonomously:
...
 
-export function buildManifestBabysitPrompt(run: RunRecord, prUrl: string): string {
+export function buildManifestBabysitPrompt(run: RunRecord, prUrl: string, rawArgs: string = prUrl): string {
+	const tokens = rawArgs.split(/\s+/);
+	const ci = tokens.includes("--ci");
+	const manifestFlagIndex = tokens.indexOf("--manifest");
+	const existingManifest = manifestFlagIndex >= 0 ? tokens[manifestFlagIndex + 1] : undefined;
+
+	const groundingStep = existingManifest
+		? `2. Use the existing manifest at '${existingManifest}' as the PR's grounding (do not synthesize a new one).`
```
aviramkofman, line 393: The `--manifest` branch at `pi/extensions/manifest-dev.ts:361` tells the executor to use the supplied manifest and not synthesize a new one, but the run was registered with `plannedManifestPath` at `pi/extensions/manifest-dev.ts:238` and verification only checks `run.manifestPath` at `pi/extensions/manifest-dev.ts:426-430`. `/babysit-pr --manifest path ...` blocks with `Manifest not found` when the executor follows this prompt. Resolve the supplied manifest path and store it in the run record, or require the executor to copy/write the existing manifest to `run.manifestPath` before execution.

.agents/plugins/marketplace.json
```diff
@@ -0,0 +1,32 @@
...
+  "plugins": [
+    {
+      "name": "manifest-dev",
+      "source": {
+        "source": "local",
+        "path": "./dist/codex/plugins/manifest-dev"
+      },
+      "policy": {
+        "installation": "AVAILABLE",
+        "authentication": "NONE"
```
aviramkofman: Codex rejects the marketplace file because `.agents/plugins/marketplace.json:15` uses `authentication: "NONE"`; `codex plugin marketplace add` reports `unknown variant NONE, expected ON_INSTALL or ON_USE`, and the second plugin repeats the same value at line 27. Replace both policy values with the current schema-supported authentication variant, or the schema-supported no-auth representation if Codex exposes one, so the documented marketplace install path can register.

README.md
```diff
@@ -91,7 +93,7 @@ Uninstall uses the same entrypoints:
 
 ```bash
 curl -fsSL https://raw.githubusercontent.com/doodledood/manifest-dev/main/dist/opencode/install.sh | bash -s -- uninstall
-curl -fsSL https://raw.githubusercontent.com/doodledood/manifest-dev/main/dist/codex/install.sh | bash -s -- uninstall
+codex plugin remove manifest-dev && codex plugin remove manifest-dev-tools && codex plugin marketplace remove manifest-dev
```
aviramkofman: The uninstall command at `README.md:96` omits the marketplace selector. Current Codex requires `PLUGIN@MARKETPLACE` or `--marketplace`, so `codex plugin remove manifest-dev` does not target the `manifest-dev` marketplace installed above. Change the removes to `codex plugin remove manifest-dev@manifest-dev` and `codex plugin remove manifest-dev-tools@manifest-dev`, and mirror the same fix in the generated Codex README/reference.

packages/manifest-dev-pi-tools/pi/extensions/manifest-dev-tools.ts
```diff
@@ -0,0 +1,30 @@
...
+	wireRuntimeHooks,
+	type ManifestCommand,
+} from "../../../../pi/extensions/manifest-dev.ts";
+
+/** Commands owned by the tools (`@doodledood/manifest-dev-pi-tools`) package. */
+const TOOLS_COMMANDS: ReadonlySet<ManifestCommand> = new Set<ManifestCommand>(["babysit-pr"]);
+
+export default function manifestDevToolsExtension(pi: ExtensionAPI): void {
+	const state = createRuntimeState();
+	// The core package owns verifier flag registration; the tools runtime reads
```
aviramkofman: The tools extension reads verifier flags during `/babysit-pr`, but `packages/manifest-dev-pi-tools/pi/extensions/manifest-dev-tools.ts:18-22` never registers them for this extension. Pi's `getFlag` only returns values for flags registered by the current extension, so `/babysit-pr --manifest-verifier-max-concurrent 1 ...` falls back to the default 24. Register the verifier flags in `manifestDevToolsExtension` before wiring the hooks, or pass `/babysit-pr` a config source that is registered on this extension.

pi/extensions/manifest-dev.ts
doodledood: Fixed in `bc54405`: `startWrapper` now parses `--manifest <path>` and sets `run.manifestPath` to the resolved supplied path (instead of the planned path), so runtime verification reads the real file. The babysit prompt header also switches to "Manifest path (existing)" in that case.

---

.agents/plugins/marketplace.json
doodledood: Fixed in `bc54405`: both entries now use `"authentication": "ON_INSTALL"` (the schema-valid variant; `NONE` is rejected). Same in the `codex-cli.md` marketplace example.

---

README.md
doodledood: Fixed in `bc54405`: uninstall now uses the `PLUGIN@MARKETPLACE` selector — `codex plugin remove manifest-dev@manifest-dev && codex plugin remove manifest-dev-tools@manifest-dev && codex plugin marketplace remove manifest-dev` — in the README, `codex-cli.md`, and `dist/codex/README.md`.

---

packages/manifest-dev-pi-tools/pi/extensions/manifest-dev-tools.ts
doodledood: Fixed in `bc54405`: `manifestDevToolsExtension` now calls `registerVerifierFlags(pi)` before wiring hooks, so the tools extension registers its own `--manifest-verifier-*` flags and `/babysit-pr` honors the overrides (Pi's `getFlag` is per-extension). Added a test asserting the flags are registered.

---

packages/manifest-dev-pi-tools/pi/extensions/manifest-dev-tools.ts
```diff
@@ -0,0 +1,33 @@
...
+const TOOLS_COMMANDS: ReadonlySet<ManifestCommand> = new Set<ManifestCommand>(["babysit-pr"]);
+
+export default function manifestDevToolsExtension(pi: ExtensionAPI): void {
+	const state = createRuntimeState();
+	// Pi's getFlag only returns values for flags registered by THIS extension, so
+	// the tools extension must register the verifier flags itself for /babysit-pr
+	// to honor --manifest-verifier-* overrides (it would otherwise fall back to
+	// defaults). Then wire its own hooks, scoped to babysit-pr runs so it never
+	// double-verifies the core package's /do or /auto runs.
+	registerVerifierFlags(pi);
```
aviramkofman, line 26: Registering the verifier flags at `packages/manifest-dev-pi-tools/pi/extensions/manifest-dev-tools.ts:24` duplicates the same three flags that `pi/extensions/manifest-dev.ts` registers in the normal repo-root Pi install. `package.json` loads both extension files, and `pi -e /tmp/manifest-dev-pr183 --help` lists every `--manifest-verifier-*` flag twice. Keep a single public flag owner, or pass the parsed values to the tools runtime without re-registering the same flag names from a second extension.

doodledood: Fixed in `b0e9f3c`: `registerVerifierFlags` now has a **single public owner**. The first extension to call it (core, listed first in the repo-root `pi.extensions`) publishes the three flags; subsequent calls are de-duplicating no-ops, so `pi --help` lists each `--manifest-verifier-*` flag once. Since Pi's `getFlag` is per-extension, `resolveVerifierConfig` reads the calling extension's own flag and falls back to the owner — so `/babysit-pr` (tools) still honors the overrides parsed by the core owner, without re-registering. A standalone tools install registers the flags itself as the owner.

Added `tests/pi_extension_flag_ownership.test.mjs` asserting each verifier flag registers exactly once when both extensions load. Bumped Pi packages 0.8.0 → 0.8.1 (+ `pi-cli.md` version/owner note).

---

pi/extensions/manifest-dev.ts
```diff
@@ -66,21 +65,30 @@ const OUTCOME_ENTRY = "manifest-dev:outcome";
...
 const FLAG_TIMEOUT_MS = "manifest-verifier-timeout-ms";
 const FLAG_MAX_CONCURRENT = "manifest-verifier-max-concurrent";
+
+// The verifier flags are process-global Pi launch flags. In the repo-root install
+// both the core and tools extensions load (and share this module), so registering
+// the same flag names from both would make `pi --help` list each flag twice. The
+// first extension to register becomes the single public owner; the other reads the
+// parsed launch value through it, since Pi's getFlag only returns values for flags
+// the calling extension itself registered.
+let verifierFlagOwner: ExtensionAPI | undefined;
```
aviramkofman: This owner is only shared in the Node test harness. Pi loads each extension with a fresh module instance, so the core extension and `manifest-dev-tools.ts` each evaluate `pi/extensions/manifest-dev.ts:85` separately; both calls see `verifierFlagOwner === undefined` and register all three flags. `HOME=$(mktemp -d) pi -e /tmp/manifest-dev-pr183 --help` still prints every `--manifest-verifier-*` flag twice on this head. Move the ownership outside this module instance, or avoid registering the tools flags when the repo-root package loads both extensions, and cover it with a real Pi loader/help-path test.

doodledood: You're right — thanks for the empirical check. The module-level owner guard isn't shared across Pi's fresh per-extension module instances, so both registered and `--help` still doubled. Fixed properly in `2e770a0` by making the single owner **static** rather than runtime state:

- **Only the core extension calls `registerVerifierFlags`; the tools extension never registers them.** That's a compile-time decision, so it holds no matter how Pi instances the modules — each flag is published exactly once.
- `/babysit-pr` still honors `--manifest-verifier-*` because `resolveVerifierConfig` falls back to reading the value from `process.argv` (process-global, shared across instances) when the tools extension's `getFlag` doesn't have it — the "pass the parsed values without re-registering" path.

Tests assert the tools extension registers **zero** verifier flags and core registers each exactly once (combined: 3, not 6), plus `launchFlagFromArgv` covers `--name value` and `--name=value`.

On the real `pi --help` subprocess test: I can't run it in this environment (no `pi` binary / `node_modules` here — the Node suite runs the extensions directly via `--experimental-strip-types`). The duplication is now eliminated by construction (the tools extension makes no `registerFlag` call for these names at all), rather than by shared runtime state, so it isn't instance-dependent. If you'd like, I can add a help-path smoke test guarded on `pi` being installed so it runs in CI but skips locally — say the word.

---

packages/manifest-dev-pi-tools/pi/extensions/manifest-dev-tools.ts
```diff
@@ -0,0 +1,33 @@
...
+/** Commands owned by the tools (`@doodledood/manifest-dev-pi-tools`) package. */
+const TOOLS_COMMANDS: ReadonlySet<ManifestCommand> = new Set<ManifestCommand>(["babysit-pr"]);
+
+export default function manifestDevToolsExtension(pi: ExtensionAPI): void {
+	const state = createRuntimeState();
+	// The tools extension does NOT register the --manifest-verifier-* flags: the core
+	// extension is their single public owner, so registering here would list each flag
+	// twice in `pi --help` (Pi loads each extension as a fresh module instance, so the
+	// two can't share an in-module owner guard). /babysit-pr still honors the overrides
+	// because resolveVerifierConfig reads the launch values from process.argv when this
```
aviramkofman: The `process.argv` fallback at `packages/manifest-dev-pi-tools/pi/extensions/manifest-dev-tools.ts:22` only works in the repo-root load, where the core extension has registered these flags. The tools package still exposes its own `pi.extensions`, but `pi -e packages/manifest-dev-pi-tools --manifest-verifier-max-concurrent 3 --print hi` exits with `Error: Unknown option: --manifest-verifier-max-concurrent` before `/babysit-pr` can run. Keep a flag owner in that load path, or remove/document tools-only flag support instead of saying `/babysit-pr` still honors the overrides.

doodledood: Good catch — the static "only core registers" approach broke exactly that standalone path. Fixed in `9fc7dea` by going with your round-4 suggestion to move ownership **outside the module instance**: a `globalThis` marker, which (unlike module-level state) *is* shared across Pi's fresh per-extension module instances in the same process. Both extensions call `registerVerifierFlags`, but the first to load claims the marker and registers; the other skips:

- **Repo-root install** (both load, core first): core owns → `pi --help` lists each flag once; tools reads values via the `process.argv` fallback.
- **Standalone tools install** (`pi -e packages/manifest-dev-pi-tools`): tools is the only loader, owns the flags, and registers them — so Pi accepts `--manifest-verifier-*` instead of rejecting them as unknown.

So `/babysit-pr` honors the overrides in **both** load paths now, and there's no double `--help` entry in the repo-root one. Tests cover both paths (resetting the `globalThis` marker between them) plus `launchFlagFromArgv`. Pi packages → 0.8.3; `pi-cli.md` ownership note updated.

The one residual: this relies on `globalThis` being shared across the two extensions' module instances (same process/realm). If your loader isolates extensions in separate realms, the marker wouldn't share and the repo-root `--help` would double again — worth a quick `pi -e <repo-root> --help` confirmation on your side. If that's the case, tell me and I'll switch to making the tools package non-standalone (drop its own `pi.extensions`) so only the repo-root path exists.

---

pi/extensions/manifest-dev.ts
```diff
@@ -92,21 +92,44 @@ export interface RuntimeState {
...
+ *   - repo-root install (both extensions load): core is listed first and owns the flags;
+ *     tools skips, so `pi --help` lists each flag once.
+ *   - standalone tools install (`pi -e packages/manifest-dev-pi-tools`): tools is the
+ *     only loader and owns the flags, so Pi accepts `--manifest-verifier-*` overrides.
+ * The non-owning extension still reads the parsed value via resolveVerifierConfig's
+ * process.argv fallback (Pi's getFlag is per-extension).
+ */
+export function registerVerifierFlags(pi: ExtensionAPI): void {
+	const slot = globalThis as Record<symbol, boolean | undefined>;
+	if (slot[VERIFIER_FLAGS_REGISTERED]) return;
```
aviramkofman, line 127: The process-wide owner bit at `pi/extensions/manifest-dev.ts:127` survives past the extension load it was meant to de-dupe. Pi creates a fresh runtime on each `loadExtensions()` call, but the `globalThis` marker remains set, so a second load in the same process registers no verifier flags at all; I reproduced `loadExtensions([core, tools])` twice and got the three flags on load 1, then `[]` on load 2. Scope the marker to one load cycle, or clear it after the paired extension load completes, so reloads/second SDK service creation still publish `--manifest-verifier-*`.

PR #184, opened 2026-06-08
Zero-agent rework: review-code skill, Codex plugin-native, Pi command rename + two-package split

doodledood:
Re-opens the work previously merged as #183 (squash `6540ab7`), which was reverted on `main` by `88a7a5c`. This pass adds a second round of review fixes (7 comments), all **anchored in the real tool sources** (npm tarballs of `@earendil-works/pi-coding-agent` and `@gotgenes/pi-subagents`; `dist/opencode/install_helpers.py`).

## What & why

Eliminates the Pi skill-leak at its root and consolidates the reviewer architecture into skills. **manifest-dev now ships zero agents** — every former agent is a skill; `verify.agent` is removed from the manifest schema. The 13 reviewers become the progressive-disclosure `review-code` skill (one dimension per invocation); functional agents become verb-named skills (`check-pr`, `poll-slack`, `review-prompt`). Codex migrates to a plugin-native marketplace distribution (installer retired). Pi commands drop the `manifest-` prefix and the Pi runtime splits into two packages.

## Pi runtime (grounded in real Pi source)

- **Two packages**: `@doodledood/manifest-dev-pi` (core: `/do`, `/auto` + shared runtime) and `@doodledood/manifest-dev-pi-tools` (`/babysit-pr`). The **tools package is repo-root-only** (no own `pi.extensions`) — loaded solely from the repo-root manifest, never standalone (which couldn't resolve its relatively-imported core runtime from its own tarball).
- **Verifier flags**: Pi's `loader.js` shows `getFlag` is gated on the calling extension and `runtime.flagValues` is rebuilt per `loadExtensions()`. So **only core registers** the `--manifest-verifier-*` flags (unconditionally — no marker that a later load cycle would wrongly suppress); `/babysit-pr` reads them via a `process.argv` fallback.
- **Verifier-spawn failures**: `@gotgenes/pi-subagents` `service-adapter.ts` reads the service's stored `currentCtx` (captured at `session_start`) for the parent snapshot; Pi's stale-context guard throws if it was invalidated. No retry (a tick can't refresh that ctx) — a systemic spawn failure becomes **one** harness-orchestration BLOCKED naming the cause + session ids.

## Skills & distributions

- **Per-target skill-activation naming**: source uses the canonical `manifest-dev:<skill>` colon form (Claude-native; OpenCode's installer rewrites to suffixed names). The **Pi dist strips the qualifier** to the bare `/skill:<name>` form Pi invokes; Codex keeps it for `$skillname`/implicit activation.
- **`/babysit-pr --ci`** now honors its one-shot contract: `--ci` is persisted on the run, and a lifecycle verification that FAILs only on external waits (`check-pr` emits a `WAIT-PENDING` marker) **exits pending** (resumable) instead of looping repair; genuinely fixable failures still repair.
- **Codex** plugin-native (`.codex-plugin/plugin.json` + repo-root `.agents/plugins/marketplace.json`, `authentication: ON_INSTALL`); namespace-metadata generation skipped for Codex. **OpenCode** stays installer-based (`AGENTS.md` context). `sync-claude-code-plugins` denylists the retired `prompt-reviewer.md` agent.

## Breaking changes
- Reviewers are addressed via the `review-code` skill dimension (`/define` encodes this); `verify.agent` is removed.
- Codex installs via the plugin marketplace (two-step).
- Pi commands dropped the `manifest-` prefix.

See `CHANGELOG.md` for migration detail.

## Versions
- `manifest-dev` plugin: 2.4.0 → 2.8.2
- `manifest-dev-tools` plugin: 0.19.0 → 0.23.1
- `@doodledood/manifest-dev-pi` (and new `@doodledood/manifest-dev-pi-tools`): 0.8.7

## Verification
- `ruff` / `black --check` / `mypy` — clean
- `pytest tests/` — 33 passed
- `node --test` Pi runtime — 43 passed

---
_Generated by [Claude Code](https://claude.ai/code/session_01VpoKL9Qz4h5M8qQUxFABeJ)_


pi/extensions/manifest-dev.ts
```diff
@@ -92,21 +92,44 @@ export interface RuntimeState {
...
+ *   - repo-root install (both extensions load): core is listed first and owns the flags;
+ *     tools skips, so `pi --help` lists each flag once.
+ *   - standalone tools install (`pi -e packages/manifest-dev-pi-tools`): tools is the
+ *     only loader and owns the flags, so Pi accepts `--manifest-verifier-*` overrides.
+ * The non-owning extension still reads the parsed value via resolveVerifierConfig's
+ * process.argv fallback (Pi's getFlag is per-extension).
+ */
+export function registerVerifierFlags(pi: ExtensionAPI): void {
+	const slot = globalThis as Record<symbol, boolean | undefined>;
+	if (slot[VERIFIER_FLAGS_REGISTERED]) return;
```
aviramkofman: The verifier flag ownership marker outlives the extension load it is de-duping. `registerVerifierFlags` stores `VERIFIER_FLAGS_REGISTERED` on `globalThis` at `pi/extensions/manifest-dev.ts:112` and returns on it at `pi/extensions/manifest-dev.ts:127-128`; Pi creates a fresh runtime for each `loadExtensions()` call in the same process, so a second core+tools load registers none of the `--manifest-verifier-*` flags. Scope the marker to one load cycle, or clear it after the paired extension load finishes.

claude-plugins/manifest-dev/skills/define/tasks/PR_LIFECYCLE.md
```diff
@@ -6,18 +6,18 @@ The goal of /do under PR_LIFECYCLE is to drive the PR to a **mergeable** state 
...
+|--------|----------|-----------|
+| PR lifecycle | `review-code`-style general-purpose agent activating the `check-pr` skill | PASS |
 
 **Templated AC** — /define synthesizes one AC per repo with the following shape:
 
 ```yaml
 verify:
-  agent: github-pr-lifecycle
   prompt: |
+    Activate the manifest-dev check-pr skill.
```
aviramkofman: OpenCode installs these skills under suffixed names, but these prompts keep prose activations that the installer does not rewrite. `claude-plugins/manifest-dev/skills/define/tasks/PR_LIFECYCLE.md:20` emits `Activate the manifest-dev check-pr skill`, and `dist/opencode/skills/review-pr/SKILL.md:43` still says to activate `review-prompt`; the installer only rewrites colon-qualified `manifest-dev:<skill>` / `manifest-dev-tools:<skill>` references at `dist/opencode/install_helpers.py:206-214`, so the installed prompts still point at missing bare skills while `dist/opencode/component-namespaces.json:11` and `dist/opencode/component-namespaces.json:23` install `check-pr-manifest-dev` and `review-prompt-manifest-dev-tools`. Generate the installed OpenCode names for these activations or extend the installer rewrite to cover this prose form.

packages/manifest-dev-pi-tools/pi/extensions/manifest-dev-tools.ts
```diff
@@ -0,0 +1,35 @@
...
+// (@doodledood/manifest-dev-pi, declared in package.json dependencies). Pi
+// installs manifest-dev as the whole repo, so the in-repo relative path is the
+// resolved location of the dependency.
+import {
+	createRuntimeState,
+	registerVerifierFlags,
+	startWrapper,
+	wireRuntimeHooks,
+	type ManifestCommand,
+} from "../../../../pi/extensions/manifest-dev.ts";
```
aviramkofman, line 10: The standalone Pi tools package imports a file that is not in its package tarball. `packages/manifest-dev-pi-tools/pi/extensions/manifest-dev-tools.ts:12` imports `../../../../pi/extensions/manifest-dev.ts`, and `packages/manifest-dev-pi-tools/package.json:19` declares `@doodledood/manifest-dev-pi`, but `npm pack --dry-run --workspace packages/manifest-dev-pi-tools` includes only `package.json` and `pi/extensions/manifest-dev-tools.ts`, so installing this workspace alone cannot resolve the shared runtime. Import the shared runtime through the declared dependency, or include the runtime files in the tools package.

pi/extensions/manifest-dev.ts
```diff
@@ -333,18 +383,30 @@ Follow the manifest-dev chain autonomously:
...
+	const tokens = rawArgs.split(/\s+/);
+	const ci = tokens.includes("--ci");
+	const manifestFlagIndex = tokens.indexOf("--manifest");
+	const existingManifest = manifestFlagIndex >= 0 ? tokens[manifestFlagIndex + 1] : undefined;
+
+	const groundingStep = existingManifest
+		? `2. Use the existing manifest at '${existingManifest}' as the PR's grounding (do not synthesize a new one).`
+		: `2. Invoke the define skill (/skill:define) with '--babysit ${prUrl} --autonomous' and write the lifecycle manifest exactly at: ${run.manifestPath}`;
+	const cadenceStep = ci
+		? `3. CI one-shot mode (--ci): perform every immediately actionable lifecycle step (inspect CI, review threads, mergeability, description sync; apply and push trusted fixes; retrigger; reply/resolve), then STOP and report the pending/waiting state instead of sleeping the runner. Do not block on long waits.`
```
aviramkofman: `--ci` changes the babysit prompt but not the runtime routing. `buildManifestBabysitPrompt` tells CI mode to stop and report wait-only state at `pi/extensions/manifest-dev.ts:395-396`, while `routeVerificationResult` sends every failed verification back into repair at `pi/extensions/manifest-dev.ts:622-628`; `check-pr` documents wait-in-progress as `FAIL` at `claude-plugins/manifest-dev/skills/check-pr/SKILL.md:79-88`. Persist a CI/no-wait run mode and route wait-only check-pr failures to the pending summary path instead of repair.

.claude/skills/sync-tools/SKILL.md
```diff
@@ -29,11 +29,9 @@ Sync only these source payloads:
...
-
-Regenerate `component-namespaces.json` on every sync from the discovered source components; never hand-maintain it. The file is part of the generated dist contract and must include every distributed skill, agent, and command. If a component exists under `dist/{cli}/skills`, `agents`, or `commands` and is missing from metadata, the sync is incomplete.
+**Namespacing model is per-CLI** (see each reference file):
+- **Installer targets (OpenCode)**: each component carries plugin ownership in `component-namespaces.json`, and install-time namespacing comes from that metadata (`-manifest-dev` / `-manifest-dev-tools` suffixes), not a single global suffix. Regenerate `component-namespaces.json` on every sync from the discovered source components; never hand-maintain it. If a component exists under `dist/{cli}/skills`, `agents`, or `commands` and is missing from metadata, the sync is incomplete.
+- **Plugin-native targets (Codex)**: the plugin is the namespace. Two plugins (`manifest-dev`, `manifest-dev-tools`) each bundle their skills under original names. No install-time suffixing, no `component-namespaces.json`, no installer.
```
aviramkofman, line 34: The Codex sync prompt still has conflicting namespace-metadata instructions after this PR made Codex plugin-native. `.claude/skills/sync-tools/SKILL.md:34` says Codex has no `component-namespaces.json`, but `.claude/skills/sync-tools/SKILL.md:67` and `.claude/skills/sync-tools/SKILL.md:85` still require regenerating `dist/{cli}/component-namespaces.json` for every CLI. Gate namespace metadata generation by target and skip it for Codex so `/sync-tools codex` does not bring back the deleted installer metadata.

.claude/skills/sync-claude-code-plugins/SKILL.md
```diff
@@ -66,7 +66,8 @@ After each sync, ensure `.agents/skills/<name>` is a symlink to `../../.claude/s
 ## Gotchas
 
 - **Source must exist**: missing source path means abort, not "delete all tracked items."
-- **Nested skills directory**: source skills live at `skills/prompt-engineering/`, `skills/review-prompt/`, etc. Copy each skill directory into `.claude/skills/<skill-name>/` — don't copy the outer `skills/` folder or you get `.claude/skills/skills/`.
+- **Nested skills directory**: source skills live at `skills/prompt-engineering/`, `skills/compress-prompt/`, etc. Copy each skill directory into `.claude/skills/<skill-name>/` — don't copy the outer `skills/` folder or you get `.claude/skills/skills/`.
+- **`review-prompt` is plugin-owned, not upstream**: manifest-dev-tools ships its own `review-prompt` skill, symlinked at `.claude/skills/review-prompt`. The upstream prompt-engineering plugin also ships a `review-prompt`, but the "skip if target is a symlink" rule above means this sync leaves the plugin's symlink alone (it neither overwrites nor deletes it). Do not re-add `review-prompt` to the tracking file.
```
aviramkofman, line 72: This new `review-prompt` gotcha only protects the skill symlink, but the sync still treats the retired `prompt-reviewer.md` agent as copyable. `.claude/skills/sync-claude-code-plugins/SKILL.md:52` copies every upstream agent and `.claude/.claude-code-plugins-sync.json:6` still tracks `prompt-reviewer.md`, so the next sync from an upstream checkout that still has that file will recreate `.claude/agents/prompt-reviewer.md`. Add an explicit retirement/denylist for `prompt-reviewer.md` and remove it from the tracked set.

dist/pi/skills/define/SKILL.md
```diff
@@ -24,12 +24,23 @@ Encode the conversation's shared understanding as a Manifest at `~/.manifest-dev
...
 
 **Verifier prompt discipline.** Before writing `verify.prompt` fields, invoke the prompt-engineering skill if it is available; if not, apply its core discipline inline. Verifier prompts are prompts: state the verifier's goal, evidence to inspect, PASS/FAIL/BLOCKED contract, and non-obvious context. Do not run a separate prompt-engineering interview — /define owns the manifest interview.
 
+**Encoding gates — always general-purpose + a skill.** There is no `verify.agent` field. Every gate is verified by a **general-purpose** subagent driven by `verify.prompt`; when a gate needs specialized behavior, the prompt tells that agent to **activate a skill**. Code-quality gates activate the `review-code` skill with a dimension; other specialized checks activate their skill (`check-pr`, `review-prompt`). Pattern:
+
+```yaml
+verify:
+  prompt: |
+    Activate the manifest-dev:review-code skill with dimension=<dimension> and review the change.
```
aviramkofman: Pi-distributed verifier prompts still use Claude plugin-qualified skill IDs. `dist/pi/skills/define/SKILL.md:38` tells verifiers to activate `manifest-dev:review-code`, and `dist/pi/skills/review-pr/SKILL.md:43` repeats that form; `dist/pi/component-namespaces.json:4` ships `review-code` as a Pi skill and `dist/pi/README.md:61` says Pi exposes installed skills as `/skill:<name>`, so these prompts do not name the installed `review-code` skill. Strip plugin qualifiers or map them to Pi's `/skill:review-code` form during Pi sync.

pi/extensions/manifest-dev.ts
doodledood: Fixed in `81b9884`, and grounded in Pi's real `loader.js` (npm tarball) rather than assumptions. `getFlag` is gated on `extension.flags.has(name)`, values live in `runtime.flagValues`, and `runtime` is rebuilt per `loadExtensions()` — so a `globalThis` marker would indeed suppress registration on a later load cycle, exactly as you said.

Resolved by dropping the marker entirely: the tools package is now **repo-root-only** (no own `pi.extensions`, see #3), so core always loads alongside and is the sole, unconditional registrant of the `--manifest-verifier-*` flags. `/babysit-pr` reads the values via the `process.argv` fallback (its `getFlag` is gated out, which is correct). Test added asserting a second core load still registers (no persisted marker).

---

packages/manifest-dev-pi-tools/pi/extensions/manifest-dev-tools.ts
doodledood: Fixed in `81b9884` by removing the standalone path rather than papering over it: the tools `package.json` no longer declares its own `pi.extensions`, so it is never installed as a standalone Pi package (which couldn't resolve the relatively-imported core runtime from its own tarball). It loads solely from the repo-root `pi.extensions`, which lists the tools extension file directly — where the relative import resolves and core is always present. Test updated to assert `"pi" not in tools` and that the repo-root manifest lists the tools extension.

---

.claude/skills/sync-tools/SKILL.md
doodledood: Fixed in `81b9884`: both the Per-CLI step 5 and the Namespace-metadata table row in `sync-tools/SKILL.md` now gate `component-namespaces.json` generation to installer/package targets (OpenCode, Pi) and explicitly skip plugin-native Codex, matching line 34 — so `/sync-tools codex` won't resurrect the retired installer metadata.

---

.claude/skills/sync-claude-code-plugins/SKILL.md
doodledood: Fixed in `81b9884`: dropped `prompt-reviewer.md` from the tracked set in `.claude/.claude-code-plugins-sync.json` and added an explicit **retirement denylist** to `sync-claude-code-plugins/SKILL.md` — the copy step now skips retired items even when an older upstream checkout still ships them, and the tracked set is refreshed minus the denylist. `.claude/agents/prompt-reviewer.md` was already removed; it can no longer be reintroduced.

---

claude-plugins/manifest-dev/skills/define/tasks/PR_LIFECYCLE.md
doodledood: Fixed in `7b5b702`. Root cause confirmed from `dist/opencode/install_helpers.py`: the rewriter only matches the plugin-qualified colon form (`manifest-dev:<skill>`), so the space form (`manifest-dev check-pr`) and bare `review-prompt` slipped through. Canonicalized the **source** to the colon form (`manifest-dev:check-pr`, `manifest-dev-tools:review-prompt`) so the existing installer rewrite catches them and the installed OpenCode prompts name `check-pr-manifest-dev` / `review-prompt-manifest-dev-tools`. Added a test pinning the source/OpenCode colon form.

---

dist/pi/skills/define/SKILL.md
doodledood: Fixed in `7b5b702`, grounded in the real Pi source: `dist/core/skills.d.ts` / `agent-session.d.ts` show skills are invoked as `/skill:<name>` with no plugin namespace, so `manifest-dev:review-code` doesn't resolve in Pi. The Pi sync now **strips the plugin qualifier to the bare name** — applied to the verifier-activation refs (`review-code`, `check-pr`, `review-prompt`) and the chain handoff (`manifest-dev:figure-out` → `figure-out`) in `dist/pi/skills/**`. Added a `pi-cli.md` conversion rule and a test asserting no `manifest-dev(-tools):<skill>` ids survive in `dist/pi`. (OpenCode keeps the colon form for its install-time suffixing; Codex keeps it for `$skillname`/implicit activation.)

---

pi/extensions/manifest-dev.ts
doodledood: Fixed in `48f37bc`. `--ci` now drives runtime routing, not just the prompt:

- `ciOneShot` is persisted on the run (parsed from `--ci` in `startWrapper`).
- The `check-pr` skill emits a stable `WAIT-PENDING` marker when the *only* remaining blockers are external waits (reviewer/CI/merge window), and the `--ci` babysit prompt instructs the lifecycle verifier to surface it.
- `routeVerificationResult`: in `ciOneShot` mode, a failure whose every FAIL gate carries `WAIT-PENDING` exits **pending** (a resumable outcome + a "pending, not blocked" status summary — "Re-run /babysit-pr --ci once the wait clears") instead of injecting repair. A genuinely fixable failure (or a mix) still routes to repair, so the no-wait/actionable distinction is preserved.

Tests cover marker detection, the pending summary, CI-pending vs CI-repair routing, and the `--ci` prompt instruction.

---

tests/test_dist_skill_references.py
```diff
@@ -349,66 +402,70 @@ def test_pi_dist_contains_only_compatible_skill_set() -> None:
...
+    """Verifier-activation / chain prose must name skills in each target's own form:
+    source + OpenCode use the canonical plugin-qualified colon form (Claude native;
+    OpenCode's installer suffixes it), while Pi strips the qualifier to the bare
+    /skill:<name> form. Anchored in Pi loader/skills source and install_helpers.py."""
+    qualified = re.compile(r"manifest-dev(?:-tools)?:[a-z]")
 
-    for command in PI_EXTENSION_COMMANDS:
-        assert f'pi.registerCommand("{command}"' in content
+    # Source canonical form: colon-qualified, never the unrewritable space form.
+    for rel in (
```
aviramkofman, line 416: `PROMPTING.md` is outside this source-side assertion, so the new per-target naming test passes while `claude-plugins/manifest-dev/skills/define/tasks/PROMPTING.md:12` still activates bare `review-prompt`. The generated Codex/OpenCode copies now activate `manifest-dev-tools:review-prompt`; source `/define` can still emit the unqualified cross-plugin verifier, and a source-to-dist sync can copy the stale task text back out. Add `PROMPTING.md` to this source check and update the source task file to the qualified id; Pi's dist should keep the stripped bare name.

pi/extensions/manifest-dev.ts
```diff
@@ -797,6 +905,86 @@ function spawnBlockedResult(gate: ManifestGate, requestedType: string, error: un
...
+/**
+ * True if a failed verification is wait-only: every FAIL gate carries the check-pr
+ * WAIT-PENDING marker (waiting on a reviewer/CI/time, not a fixable defect). Used only
+ * in --ci one-shot mode to exit pending instead of looping repair. A mix of wait and
+ * real failures is not wait-only — repair still runs.
+ */
+export function isWaitPendingFailure(verification: VerificationRecord): boolean {
+	const failures = verification.results.filter((result) => result.verdict === "FAIL");
+	if (failures.length === 0) return false;
+	return failures.every((result) =>
```
aviramkofman: `isWaitPendingFailure()` depends on a raw `WAIT-PENDING` substring in `evidence/details`, but `claude-plugins/manifest-dev/skills/check-pr/SKILL.md:91` introduces that caller-specific token while the same prompt says `check-pr` carries no workflow-specific tokens and forbids tokens outside the fixed vocabulary. A compliant verifier can return the normal wait FAIL (`bash sleep 600; reinvoke`) without the marker, and `/babysit-pr --ci` routes the wait-only PR into repair. Make wait-pending a structured verifier contract Pi parses, then update the check-pr FAIL examples and Pi runtime reference so the marker is legal and required.

tests/test_dist_skill_references.py
doodledood: Fixed. Qualified both activations at the source `PROMPTING.md:12` (`manifest-dev:review-code`, `manifest-dev-tools:review-prompt`) and propagated per-target — Codex/OpenCode colon, Pi bare. Extended `test_skill_activation_names_are_per_target` to assert the source `PROMPTING.md` carries the qualified id and never the bare `activates the \`review-prompt\` skill` form, so a sync can't copy stale text back out.

---

pi/extensions/manifest-dev.ts
doodledood: You're right — `WAIT-PENDING` violated check-pr's workflow-neutrality invariant (it "does not carry workflow-specific tokens" / "no tokens outside the fixed vocabulary"). Per your pick (verifier→runtime layer), I kept check-pr neutral and moved the contract out of it:

- **Reverted** the `WAIT-PENDING` paragraph from check-pr (source + all 3 dist copies). check-pr stays a workflow-neutral reporter.
- The **`--ci` lifecycle verifier** now *derives* wait-only from check-pr's existing structured FAIL directives — every failing gate is a wait (`bash sleep <N>; reinvoke`) and none is actionable (no reply / reply-and-resolve / retrigger / sync-description / re-request-review / prose fix) — and emits `WAIT-PENDING` in its EVIDENCE/DETAILS. That token is the verifier→runtime contract, not check-pr's.
- Documented the contract in `pi-cli.md` (Pi runtime reference) and reworded the `WAIT_PENDING_MARKER` constant + CHANGELOG to match. `isWaitPendingFailure`/`routeVerificationResult` parse it unchanged.

So "exit pending vs repair" stays a caller decision, derived from neutral check-pr state. `48f37bc`→ latest.

---

pi/extensions/manifest-dev.ts
```diff
@@ -333,18 +378,30 @@ Follow the manifest-dev chain autonomously:
...
+		: `2. Invoke the define skill (/skill:define) with '--babysit ${prUrl} --autonomous' and write the lifecycle manifest exactly at: ${run.manifestPath}`;
+	const cadenceStep = ci
+		? `3. CI one-shot mode (--ci): perform every immediately actionable lifecycle step (inspect CI, review threads, mergeability, description sync; apply and push trusted fixes; retrigger; reply/resolve), then STOP and report the pending/waiting state instead of sleeping the runner. Do not block on long waits. The lifecycle gate's verifier owns the wait-only decision — check-pr stays workflow-neutral and emits no such token; the verifier DERIVES it from check-pr's FAIL output: if every failing gate's directive is a wait ("bash sleep <N>; reinvoke") and none is actionable (no reply / reply-and-resolve / retrigger / sync-description / re-request-review / prose fix), the PR is blocked only on external waits. In that case the verifier must return VERDICT: FAIL with the token ${WAIT_PENDING_MARKER} in its EVIDENCE/DETAILS so the runtime exits pending instead of looping repair. If any failing gate is actionable, omit ${WAIT_PENDING_MARKER} and report a normal FAIL.`
```
aviramkofman: The `WAIT-PENDING` derivation rule is only in `buildManifestBabysitPrompt`, so the verifier that must emit the marker does not receive it. `buildGateVerifierPrompt` passes only the manifest gate's `verify.prompt` as verifier instructions (`pi/extensions/manifest-dev-runtime.ts:111`), and `claude-plugins/manifest-dev/skills/define/tasks/PR_LIFECYCLE.md:20` still templates that prompt as just `Activate the manifest-dev:check-pr skill`; `--manifest` runs skip synthesis entirely. `isWaitPendingFailure` then only scans for the marker, so wait-only `check-pr` FAILs route back to repair instead of pending. Move the derivation rule into the generated lifecycle `verify.prompt`/template, or wrap `--ci` lifecycle verifier prompts in the runtime before spawning them.

doodledood: Good catch — the fix was inert. The derivation rule sat in `buildManifestBabysitPrompt` (executor prompt), but verifiers are spawned with `buildGateVerifierPrompt`, which wraps only the gate's `verify.prompt`; the lifecycle template is just `Activate the manifest-dev:check-pr skill`, and `--manifest` runs skip synthesis. So the verifier never saw the rule and `isWaitPendingFailure` never matched.

Fixed by moving the contract into the runtime where it reaches the verifier:
- `WAIT_PENDING_MARKER` + new `CI_WAIT_PENDING_VERIFIER_RULE` now live in `manifest-dev-runtime.ts`.
- `buildGateVerifierPrompt` takes `ciOneShot`; when set, it injects the derivation rule into the **verifier prompt** — so it reaches the verifier subagent for synthesized *and* `--manifest` runs alike. Threaded `ciOneShot: run.ciOneShot` at the spawn site.
- The executor `--ci` prompt is trimmed to the pending cadence only; it no longer names the token (the executor doesn't emit it).
- `pi-cli.md`/CHANGELOG updated; the drift guard now asserts the rule is in the verifier builder, and a node test asserts the verifier prompt carries `WAIT-PENDING` only when `ciOneShot`.

Pushed in the latest commit. Gates: node 44, pytest 37, lint/types clean.

---

pi/extensions/manifest-dev-runtime.ts
```diff
@@ -6,12 +6,22 @@ export type VerificationStatus = "passed" | "failed" | "blocked";
...
 
+// Verifier→runtime contract token (NOT a check-pr token — check-pr stays workflow-neutral).
+// In a `/babysit-pr --ci` (ciOneShot) run the runtime injects CI_WAIT_PENDING_VERIFIER_RULE
+// into every gate's verifier prompt; a verifier that finds the gate is blocked only on an
+// external wait appends this marker, and the runtime routes the failure to a pending exit
+// instead of repair. Defined here (not in the executor prompt) so it reaches the verifier
+// subagent for synthesized AND --manifest runs alike.
+export const WAIT_PENDING_MARKER = "WAIT-PENDING";
+export const CI_WAIT_PENDING_VERIFIER_RULE = `
+CI one-shot wait-pending rule: if this gate is blocked ONLY on an external wait — the check-pr finding's only directive is a wait ("bash sleep <N>; reinvoke") and nothing is actionable (no reply / reply-and-resolve / retrigger / sync-description / re-request-review / prose fix) — append the token ${WAIT_PENDING_MARKER} on its own line in DETAILS so the runtime exits pending instead of looping repair. Omit ${WAIT_PENDING_MARKER} entirely if any failing gate is actionable (report a normal FAIL).`;
```
aviramkofman: The injected wait-pending rule still points the verifier at the wrong verdict. It describes the condition as `blocked ONLY on an external wait`, and the base verifier rules at `pi/extensions/manifest-dev-runtime.ts:133-135` say external/human blockers are `BLOCKED`; but `routeVerificationResult` only takes the pending path when `verification.status === "failed"`, and `isWaitPendingFailure` only scans `FAIL` gates. A verifier that follows the prompt can return `BLOCKED` with `WAIT-PENDING`, and `/babysit-pr --ci` will go through the generic blocked/escalation path instead of the pending path. Tell the verifier to return `VERDICT: FAIL`, not `BLOCKED`, for wait-only lifecycle states, or teach the runtime to treat `BLOCKED` + `WAIT-PENDING` as pending.

doodledood: Valid catch — confirmed: `aggregateVerificationStatus` returns `"blocked"` if any gate is BLOCKED, and a reviewer/CI/merge-window wait genuinely classifies as BLOCKED under the base verdict rules (human/external, not workspace-repairable). My pending path only fired on `status === "failed"`, so `BLOCKED + WAIT-PENDING` fell through to generic escalation.

Took the runtime route (your option b) so it's robust to the verifier's verdict choice:
- `isWaitPendingFailure` → `isWaitPendingVerification`: scans every **non-PASS** gate (FAIL *or* BLOCKED) for the marker; wait-only iff ≥1 non-PASS gate and all carry `WAIT-PENDING`.
- `routeVerificationResult`: the `ciOneShot` pending branch now fires for `"failed"` **or** `"blocked"` before the repair/escalate branches.
- The injected rule now tells the verifier to pick the verdict per the base rules (a wait is *typically BLOCKED*) and append `WAIT-PENDING`; the runtime routes either verdict to pending.
- New node test asserts `BLOCKED + WAIT-PENDING` → pending, not escalation.

Pushed. Gates: node 45, pytest 37, lint/types clean.

---

tests/pi_extension_runtime.test.mjs
```diff
@@ -893,6 +942,208 @@ test("routeVerificationResult injects repair, blocked, and done runtime outcomes
...
+		// Outcome recorded as a (resumable) escalate flavored pending; run left blocked.
+		assert.equal(calls.entries.some((e) => e.customType === "manifest-dev:outcome" && e.data.summary === "CI one-shot pending on external wait."), true);
+		assert.equal(calls.entries.some((e) => e.customType === "manifest-dev:run" && e.data.status === "blocked"), true);
+	} finally {
+		rmSync(routeRunStateFile, { force: true });
+		rmSync(tmp, { recursive: true, force: true });
+	}
+});
+
+test("routeVerificationResult exits pending for a --ci wait-only BLOCKED verification (not generic escalation)", async () => {
```
aviramkofman, line 1074: `tests/pi_extension_runtime.test.mjs:1074` adds the positive `ciOneShot` BLOCKED + `WAIT-PENDING` route, but the suite only has the unmarked negative for `failed` at `tests/pi_extension_runtime.test.mjs:1117`. A route that sends every `ciOneShot` `blocked` verification to pending would still pass. Add the paired `blocked`/no-`WAIT-PENDING` case and assert the generic blocked escalation path: blocked status message, blocker outcome, and no pending summary.

doodledood: Valid gap — the positive test alone didn't pin the marker's role on the blocked path. Added the paired negative: a `ciOneShot` `blocked` verification with **no** `WAIT-PENDING` takes the generic blocked/escalation path — asserts the `is blocked for …` status message, the `"Harness verification is blocked."` outcome, **no** pending summary, and no repair. Now a route that sent every `ciOneShot` blocked verification to pending would fail this test. Pushed (test-only). Gates: node 46, pytest 37.

---

tests/pi_extension_runtime.test.mjs
```diff
@@ -893,6 +942,245 @@ test("routeVerificationResult injects repair, blocked, and done runtime outcomes
...
+		latestVerificationByRunId: new Map(),
+		activeRunByExecutorSessionId: new Map([[run.executorSessionId, run]]),
+		childSessionIds: new Set(),
+	};
+	const verification = {
+		runId: run.runId, manifestPath: "/m", manifestSha256: "h", requestedAt: "s", completedAt: "e",
+		cwd: process.cwd(), status: "blocked", orchestratorSessionId: "manifest-verify-1",
+		results: [{ ...gateResult("BLOCKED"), gateId: "AC-1.1", title: "PR lifecycle", evidence: "Needs maintainer secret to retrigger; no WAIT marker here" }],
+	};
+	await routeVerificationResult(pi, ctx, state, run, verification);
```
aviramkofman: This new route test leaves `~/.manifest-dev/runs/manifest-dev-ci-escalate.json` behind. `routeVerificationResult` writes the run-state file before branching, and the neighboring route tests define `routeRunStateFile`, remove it before the test, and clean it in `finally`; this one calls `routeVerificationResult` at `tests/pi_extension_runtime.test.mjs:1173` without that cleanup. Add the same `rmSync(..., { force: true })` before/finally cleanup for `manifest-dev-ci-escalate.json` so local and CI runs do not leave stale manifest-dev state in `$HOME`.

doodledood: Valid — confirmed `routeVerificationResult` writes `~/.manifest-dev/runs/<runId>.json` before branching (`manifest-dev.ts:608`), so the test left `manifest-dev-ci-escalate.json` behind. Added the same `rmSync` before/`finally` cleanup the neighboring route tests use. Also fixed the sibling `still repairs` test (`manifest-dev-ci-repair.json`), which had the identical latent leak. Verified the suite leaves no `manifest-dev-ci-*` state in `$HOME`. Pushed (test-only). Gates: node 46.

---

PR #185, opened 2026-06-10
slack-mrkdwn: correct list guidance, per-send-path strikethrough

Dev-iL:
Correction to the slack mrkdwn instructions, following manual testing.

### DIFF
--- a/claude-plugins/manifest-dev/skills/figure-out/references/slack-mrkdwn.md
+++ b/claude-plugins/manifest-dev/skills/figure-out/references/slack-mrkdwn.md
@@ -8,15 +8,15 @@ Slack messages render via `mrkdwn`, not GitHub-flavored markdown. The flavors di
 |--------|--------------|------------------------|
 | Bold | `*bold*` | `**bold**` |
 | Italic | `_italic_` | `*italic*` |
-| Strikethrough | `~strike~` | `~~strike~~` |
+| Strikethrough | `~strike~` — but rendering varies by send path, so verify it in your setup (some paths render `~~strike~~`) | assuming one form works on every send path |
 | Inline code | `` `code` `` | same |
 | Code block | ```` ```code``` ```` (triple backticks) | same |
 | Link with label | `<https://example.com\|label>` | `[label](https://example.com)` |
 | Bare link | `<https://example.com>` or just the URL | bare URL works too |
 | Section title | `*Title*` on its own line (bold; no header syntax) | `# Title`, `## Title` |
 | Blockquote | `> quoted line` | same |
-| Bullet list | `- item` on its own line | same (no special syntax — just hyphens + newlines) |
-| Numbered list | `1. item` on its own line | same |
+| Bullet list | `- item`, each item on its own physical line | items run together on one line — `•` or `-` mid-line is just text |
+| Numbered list | `1. item`, each item on its own physical line | relying on `1.` being parsed — there is no list parsing in mrkdwn |
 | User mention | `<@U123ABC>` | n/a |
 | Channel mention | `<#C123ABC\|name>` | n/a |
 | User group ping | `<!subteam^S123ABC>` | n/a |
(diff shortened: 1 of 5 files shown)


Dev-iL:
@doodledood (or @claude ?) can you review/merge this please?

claude-plugins/manifest-dev/skills/figure-out/references/slack-mrkdwn.md
```diff
@@ -8,15 +8,15 @@ Slack messages render via `mrkdwn`, not GitHub-flavored markdown. The flavors di
 |--------|--------------|------------------------|
 | Bold | `*bold*` | `**bold**` |
 | Italic | `_italic_` | `*italic*` |
-| Strikethrough | `~strike~` | `~~strike~~` |
+| Strikethrough | `~strike~` via the raw API or korotovsky/slack-mcp-server; `~~strike~~` via the claude.ai Slack MCP | assuming one form works on every send path |
```
doodledood: Please avoid mentioning the how (aka via raw api or specific MCP) since everyone has a different setup and it may cause issues

dist/codex/plugins/manifest-dev/skills/figure-out/references/slack-mrkdwn.md
```diff
@@ -8,15 +8,15 @@ Slack messages render via `mrkdwn`, not GitHub-flavored markdown. The flavors di
 |--------|--------------|------------------------|
 | Bold | `*bold*` | `**bold**` |
 | Italic | `_italic_` | `*italic*` |
-| Strikethrough | `~strike~` | `~~strike~~` |
+| Strikethrough | `~strike~` via the raw API or korotovsky/slack-mcp-server; `~~strike~~` via the claude.ai Slack MCP | assuming one form works on every send path |
```
doodledood: Same

dist/opencode/skills/figure-out/references/slack-mrkdwn.md
```diff
@@ -8,15 +8,15 @@ Slack messages render via `mrkdwn`, not GitHub-flavored markdown. The flavors di
 |--------|--------------|------------------------|
 | Bold | `*bold*` | `**bold**` |
 | Italic | `_italic_` | `*italic*` |
-| Strikethrough | `~strike~` | `~~strike~~` |
+| Strikethrough | `~strike~` via the raw API or korotovsky/slack-mcp-server; `~~strike~~` via the claude.ai Slack MCP | assuming one form works on every send path |
```
doodledood: ^

claude-plugins/manifest-dev/skills/figure-out/references/slack-mrkdwn.md
Dev-iL: How do you suggest to specify this otherwise? There is no single way that works. Should we tell the agent to try both and remember which one works? Or do we assume that all MCPs work one way and all APIs work the other?

Dev-iL: For now I changed it to 
> `~strike~` — but rendering varies by send path, so verify it in your setup (some paths render `~~strike~~`)

doodledood: Honestly I'm thinking maybe even remove the file completely since it should know what mrkdwn is by itself and can adapt. Llm is smart 

By specifying the exact thing we may be over fitting in a sense. 

I also checked the ref link again and slacks own docs say it uses the single ~ only


Can you try just removing the full mrkdwn file and its reference  altogether and see if it just works?


If it's not I still think we should adhere to what's in the slack link and if you are using some MCP that deviates it should live in your own CLAUDE.md as a point

Dev-iL: - The reason we created this file to begin with is because the llm did _not_ format it correctly by itself.
- It's not just "some" MCP -  the official slack integration for claude doesn't format strikethrough correctly according to my tests. I'll try to file an upstream bug report, just need to find where.

doodledood: I meant in the beginning nothing was mentioned but my intuition tells me that just mentioning mrkdwn in the figure out team skill is enough

PR #192, opened 2026-06-20
fix(figure-out): close --autonomous premature-read-naming gap

doodledood:
## Problem

`figure-out --autonomous` was naming its read after a first pass, then leaving the user to hand-crank load-bearing presses the discipline already mandates — informed-dissent/disconfirming test, outside-view base rates, decision-relevance, recency/claim-decay, and source incentives (skin-in-the-game). Each maps to a rule already in `SKILL.md`/the probe files, so the defect is **premature stopping in autonomous mode**, not missing discipline.

Two lines in `references/autonomous.md` drove it:
- A claim that "the investigation and all of its discipline are **unchanged**" — but losing the user removes a *second presser* and the one-question-per-turn cadence, so the bar should **rise**, not hold.
- A standalone "**Stop when the read is named**" that licensed name-then-stop.

## Fix (localized, anti-bloat — net +3/−2 in `autonomous.md`)

1. **Reframe line 3** — no user to *press back* ⇒ completeness bar **rises**; the agent plays both roles.
2. **Detach stop from naming** — gate it on the pre-naming presses being run; keep only the no-chain-into-execution stop.
3. **Point at existing discipline instead of enumerating** — self-supply presses *by reference* (master pre-naming gate + topic probe file), and run `SKILL.md`'s **independent re-derivation pass**, which already applies in the "no one will audit before relied on" case that autonomous mode *is*. No hardcoded probe checklist (that would be mechanism-as-prescription, and the originating session's skin-in-the-game press wasn't in any four-item list).
4. **Unify a split-rule** — the no-execution stop now lives once, in `autonomous.md`; `SKILL.md` keeps the general "naming the read ends the skill, in every mode."

## Housekeeping
- Regenerated dist copies (codex / opencode / pi) of `figure-out/references/autonomous.md` (identity transform).
- Version bumps: `manifest-dev` 2.11.0→2.11.1; Pi packages 0.11.0→0.11.1 (a `dist/pi/` shared asset changed) + opencode plugin mirror; kept `sync-tools/references/pi-cli.md`'s package-version example in sync.

## Verification
Built via `/figure-out` → `/define` → `/do`. The manifest verifies each change by subagent, plus a `review-prompt` gap-calibration gate and a `change-intent` review; `ruff`/`black`/`mypy` clean (no Python touched).

---
_Generated by [Claude Code](https://claude.ai/code/session_01LgNQXhWJX4yUgmgMCtcaCs)_


.claude/skills/sync-tools/references/opencode-cli.md
```diff
@@ -95,6 +95,6 @@ None shipped. OpenCode hooks are additional functions on the same plugin object
 1. **Frontmatter controls mostly ignored by OpenCode** — `disable-model-invocation` has no OpenCode effect, and all skills remain model-visible through the `skill` tool. `user-invocable` is honored only by manifest-dev's plugin when registering slash-command wrappers (`false` suppresses the wrapper, not the skill).
 2. **Bare names, first-found-wins** — same-name skills dedup by discovery order with a logged warning; a project-local `do` shadows manifest-dev's skill. Same-name commands are preserved rather than overwritten, so a user/project `/do` command shadows manifest-dev's slash wrapper.
 3. **$ARGUMENTS in command wrappers** — OpenCode command templates support `$ARGUMENTS`; the wrapper passes them through in the prompt `Use the <skill> skill with: $ARGUMENTS`.
-4. **No Stop-hook backstop for `/do`** — use `/goal /do <manifest-path>` for unattended continuation.
+4. **No Stop-hook backstop for `/do`** — for unattended continuation use a `/goal` whose argument is a completion condition (all Acceptance Criteria and Global Invariants PASS), as the skills print. Keep the `/goal` backstop blocks intact (OpenCode supports Codex-style goal mode); do not strip them as Pi does.
```
doodledood: It's a Claude code skill. Keep it general don't mention other tools. Also ensure the /goal is free of manifest (aka manifest stuff is in the goal prompt) because goal may work shallowly without reading files etc

doodledood: Both addressed in 49a0d96:

1. **Kept the sync note general** — `opencode-cli.md` no longer names Codex/Pi or bakes in manifest-specifics; it just says to use the self-contained `/goal` completion-condition the skills print and keep the `/goal` blocks.

2. **`/goal` is now free of the manifest** — you're right that the evaluator is a fresh model judging the *transcript* with no file reads. `/define`'s printed goal is now self-contained: it embeds a one-line `<deliverables>` summary and a transcript-observable "done" (`/do reports every gate PASS and emits /done`), instead of pointing the condition at the manifest file. The `<manifest-path>` that remains is only the executor's read target, not something the evaluator must open. `figure-out --autonomous` and `/auto`'s goals were already self-contained (Read-anatomy / full-chain milestones), and `/do`'s note now points at define's self-contained form.

Pi still drops the whole `/goal` block (no `/goal` there); Codex/OpenCode keep it. All manifest gates re-verified green.

If by "keep it general" you also meant the wording inside the distributed skills (not just this sync reference), let me know and I'll sweep those too.

---

PR #195, opened 2026-06-22
Carry the autonomous posture in /goal and make its emission reliable

doodledood:
## What & why

The prior PR (#192) turned the printed `/goal` strings into measurable **completion conditions**. This builds on that with the **autonomous operating posture** — the "clear all fog that can be cleared without a human" discipline that previously lived only inside `figure-out/references/autonomous.md` and never rode along with the goal itself. Three gaps, one change:

1. **Posture in every printed `/goal`** — figure-out `autonomous.md`, `define`'s handoff, and `auto` now append a clause to the goal: *resolve every question you can yourself, record low-confidence calls as assumptions, and only halt for a blocker that genuinely needs the user.* It's kept **distinct** from the completion condition so the fresh-model evaluator's done-test stays crisp. The three READMEs' goal examples get the same clause.

2. **Reliable emission** — `figure-out --autonomous` was observed continuing *without* printing its backstop. Its print directive is now an explicit first action, and the "don't print a nested goal under `/auto`" suppression **fails toward printing** (when it can't clearly see it's under `/auto`, it prints — a missing backstop costs more than a redundant one). `auto` previously only *documented* its goal as prose; it now actively prints the chain-complete backstop and stays the chain-owner.

3. **Proactive mode prose** — `autonomous.md` gains the standing principle to proactively clear all fog investigation or a defensible default can settle without a human, phrased as an **intensification** of the master frame's "explore instead of asking," not a restatement. The no-chain-into-execution boundary is preserved.

## Guardrails honored
- No new headings/sections in `autonomous.md`; turn bound stays a user-filled `N` placeholder.
- The fog-clearing principle intensifies rather than duplicates the master rule (avoids the split-rule trap).
- Distributions regenerated via `sync-tools`: **Pi drops** the `/goal` block, **Codex/OpenCode keep** it (with the new clause). The full sync also corrected pre-existing main drift (codex tools `0.23.1 → 0.26.0`, stale `tasks/CODING.md`) so dist matches source.
- `manifest-dev` `2.14.0 → 2.15.0` (minor); Pi packages `0.11.3 → 0.11.4` in lockstep; opencode mirror `→ 2.15.0`.

## Verification
Encoded as a manifest (`.manifest/goal-autonomous-posture-emission-2026-06-22.md`) and executed via `/do`: `review-code` change-intent (no scope creep) + `review-prompt` gap-calibration, plus anti-bloat, version-lockstep, dist byte-match, and `ruff`/`black`/`mypy` gates.

https://claude.ai/code/session_01JDEnQjMyQ1vGy7SXnTUEFi

---
_Generated by [Claude Code](https://claude.ai/code/session_01JDEnQjMyQ1vGy7SXnTUEFi)_


claude-plugins/manifest-dev/skills/figure-out/references/autonomous.md
```diff
@@ -1,14 +1,14 @@
...
 
-On activation, before the first question, print a copy-pasteable backstop for the run — a `/goal` whose argument is the Read's completion condition, so the host's fresh-model evaluator re-opens the turn if you stop early:
+Your first action on activation — before the first question, before any investigation — is to print a copy-pasteable backstop for the run: a `/goal` whose argument is the Read's completion condition, so the host's fresh-model evaluator re-opens the turn if you stop early:
 
 ```
-/goal Investigate <topic> until the Read is named with full anatomy — every load-bearing branch pressed, the independent re-derivation run, the rival set no longer moving; do not stop at a first-pass read. Stop after N turns if it stalls.
+/goal Investigate <topic> until the Read is named with full anatomy — every load-bearing branch pressed, the independent re-derivation run, the rival set no longer moving; do not stop at a first-pass read. Resolve every question you can yourself and record low-confidence calls as assumptions, halting only for a blocker that genuinely needs me. Stop after N turns if it stalls.
```
aviramkofman: This and the rest is not enough. Must clear all fog that can be cleared without intervention by user. Trustworthiness and truth is above speed here

doodledood: Strengthened — pushed in 3eb303a. The principle now mandates clearing rather than satisficing, and makes the value ordering explicit:

- **autonomous.md prose:** *"you must clear every patch of fog that investigation or derivation can clear without the user — dig to the truth rather than guessing, because here trustworthiness and truth outrank speed. An assumption is the last resort, never a shortcut around work you could have done yourself."*
- **Every printed `/goal` clause:** *"Clear all fog you can without me — investigate to the truth rather than guessing, since trustworthiness outranks speed — recording an assumption only for what genuinely needs me, and halting only for a blocker that truly requires my input."*

The key shift: an assumption is now framed as the *last resort* for genuinely-user-dependent fog, not a fast way to log-and-move-on. `review-prompt` passed it as bounded gap-closing emphasis (the "you must"/"every" is scoped to *clearable-without-the-user* fog), not an unjustified absolute.

---

PR #215, opened 2026-07-04
Surface user latent criteria in figure-out; rename Approach → Initial Approach

doodledood:
## Summary

Three first-principles additions distilled from a figure-out session on Thariq's "A Field Guide to Fable: Finding Your Unknowns." The article turned out to be ~90% validation of the existing architecture; four tempting candidates dissolved under pressure-testing, and three survived — all aimed at the one quadrant the workflows under-serve: **unknown-knowns**, the user's own latent criteria that interrogation can't extract.

## What changed

**1. figure-out — latent-criteria artifact offer.** When the model senses a taste/shape/recognition criterion the user holds but can't state ("I'll know it when I see it"), questions won't reach it — only reacting to something concrete will. figure-out now *must proactively offer* to produce a concrete artifact (a reference to point at, a quick mock, or a few divergent options), proposing the cheapest that would crack the fog. The **offer is mandatory** when the fog is sensed; **production is token-bound** and waits on the user's yes. It's a one-shot probe, explicitly distinct from scratch mode (the artifact *leads* and is consumed, vs. scratch mirroring settled understanding). In autonomous mode it degrades through the existing user-addressed offer-resolution rule: no reactor → don't produce → record the criterion as an `ASM-*`.

**2. figure-out — user-epistemic-state as a first-class variable.** figure-out modeled uncertainty about the *territory* but not the *user's* starting knowledge. It now senses the user's expertise in the specific problem/domain/codebase and calibrates blindspot depth, question pitch, and explain-vs-assume; when the user is a domain novice, teaching them mid-deliberation is part of surfacing their unknowns (distinct from the post-hoc teach-me skill). Interactive-only — inert in autonomous mode.

**3. define — `Approach` → `Initial Approach` + tacit-criteria encoding rule.** The manifest's `Approach` section was always meant as a soft suggestion (success-criteria-first was the premise); the rename makes that softness legible to `/do`. Added an encoding rule: criteria a user pinned *by reacting* to something concrete are success criteria and must be preserved (Acceptance Criterion / Global Invariant / Process Guidance), never folded into the soft Initial Approach where `/do` could silently pivot away. No new manifest category — routes onto existing structure.

## Housekeeping

- Bumped `manifest-dev` plugin `3.3.0 → 3.4.0` and Pi package `1.2.0 → 1.3.0`.
- Regenerated the OpenCode / Codex / Pi distributions under `dist/` (body-only edits, per-CLI substitutions preserved).
- Updated `CONTEXT.md`, plugin README, and the `pi-cli.md` version example for naming consistency.

## Verification

Executed via `/auto` (figure-out → define → do). Every Acceptance Criterion and Global Invariant was verified by an independent subagent against the committed diff:

- 9 content ACs — each addition's load-bearing elements present and coherent — **PASS**
- Prompt quality (review-prompt, no MEDIUM+) and change-intent (review-code, no LOW+) — **PASS**
- Version bumps, `ruff`/`black`/`mypy` clean, symlink integrity, distribution sync — **PASS**

Explicitly out of scope: `/do` deviation-logging — the manifest architecture already handles criteria changes via amendment and treats Initial-Approach path departures as unremarkable by design.

---
_Generated by [Claude Code](https://claude.ai/code/session_01X1nFFwX6XttyZ4fn9DMtqp)_


claude-plugins/manifest-dev/skills/figure-out/SKILL.md
```diff
@@ -30,6 +30,10 @@ For evidence-heavy investigations, keep a live belief register: current leading
 
 When a read implies changing or removing an existing state, behavior, constraint, or artifact, test the status quo's possible job first: why might it exist, and is that purpose still wanted? Treat status-quo intent as evidence to weigh, not a veto.
 
+Read the user's starting point, not just the topic's: how much do they already know here — fluent in this codebase, or new to the domain? Calibrate to it — how deep to run the blindspot pass, how to pitch questions, how much to explain versus assume. When the user is new to the domain, their unknowns include ones they can't recognize yet; teaching them the terrain — mid-deliberation, enough to hold a criterion — is part of surfacing those, not a detour (distinct from teach-me, which explains finished work after the fact). This reads the user; with no user to read — autonomous or unattended — it's inert.
```
doodledood: I worry this will cause every session to begin with questioning the user about their level. It should be more subtle and understand it by itself based on the user input / past / how they frame themselves and adapt

doodledood: Fixed in 68ac3b1. Reworded so the model **infers** the user's level from how they show up — framing, vocabulary, what they take as given, any experience they mention — and keeps adjusting as they reveal more, with an explicit "not by quizzing them on their level." No opening question about expertise; it stays passive and adaptive.

---

PR #218, opened 2026-07-05
Fix skill activation wording and branch race guidance

Dev-iL:
## Summary

Closes two instruction-following gaps found while investigating `/do` verifier behavior:

1. **Skill-activation wording.** `verify.prompt` templates that told a verifier to "Activate the manifest-dev:X skill…" produced inconsistent literal skill invocation — verifiers would sometimes free-hand the check with their own ad hoc inspection instead of actually loading the named skill's rubric. Sampled evidence showed compliance was high when the review instructions stood alone but dropped sharply once embedded inside a larger manifest (the realistic shape of real usage), and one sampled skip changed a verdict outright on identical underlying code.
2. **Shared-worktree branch-mutation race.** `/do`'s default phase is 1 for every criterion unless the author sets `phase:` explicitly, and same-phase criteria launch in parallel sharing the orchestrator's working directory. A verifier that needs a specific branch's actual checked-out state and runs `git checkout <branch>` in place mutates that shared directory while sibling criteria concurrently read it.

Both are prompt/guidance edits — no executable-code changes.

## Changes

- **`claude-plugins/manifest-dev/skills/define/SKILL.md`**, **`define/tasks/CODING.md`**, **`define/tasks/PR_LIFECYCLE.md`** — reworded the four `verify.prompt`-embedded template strings from "Activate the manifest-dev:X skill…" to an explicit anti-free-hand directive: *"Invoke the manifest-dev:X skill for real…; do not free-hand this … instead of actually invoking the skill."* Deliberately avoids naming a Claude-Code-specific "Skill tool" primitive, since these templates are distributed by `sync-tools` into OpenCode, Codex, and Pi packages that have no conversion rule for that phrase. The nested-spawn-avoidance warning and PASS/FAIL/BLOCKED contract are unchanged in meaning.
- **`define/tasks/CODING.md`** (Defaults) — added "Isolate branch-specific verifiers": when a `verify.prompt` needs a branch's actual checked-out state, isolate via a disposable `git worktree add <tmp-dir> <branch>` / `git worktree remove --force <tmp-dir>` rather than an in-place `git checkout`; fall back to giving the criterion its own `phase:` when a dedicated worktree isn't practical (including when a sibling already holds that branch checked out — `git worktree add` fails outright, not just races, in that case).
- **`do/SKILL.md`** — added a matching "Shared-worktree hazard" note on the execution side, cross-referencing the CODING.md Default.
- **`docs/adr/20260704-require-explicit-skill-tool-invocation-in-verify-prompts.md`** and **`docs/adr/20260704-isolate-branch-specific-verifiers-via-disposable-worktrees.md`** — new ADRs recording both decisions, cross-linked via `Related:`. Written in self-contained prose (no references to the originating investigation's internal run artifacts, session IDs, or criterion codes) so they read standalone on `main`.
- **`claude-plugins/manifest-dev/.claude-plugin/plugin.json`** — patch bump 3.4.1 → 3.4.2.

## Trade-offs

- Worktree isolation is the primary fix for the branch-mutation race; phase separation is the documented fallback for tooling that must run from the repo's exact original path, or when a sibling already holds the same branch's worktree — encoding both rather than picking one, per explicit review-code findings that each option alone left a gap.
- The first draft of the wording fix explicitly named "the Skill tool" for maximum clarity on Claude Code, but that was reworded to a host-agnostic forcing phrase after review flagged it as a harness-specific primitive baked into distributed template text (see the D1 ADR's Alternatives Considered).

## Verification

Each Acceptance Criterion / Global Invariant was independently verified by a general-purpose subagent running the manifest's own `verify.prompt` (review-code across several dimensions, review-prompt, and direct file/version checks). All gates have fresh PASS evidence as of the branch's current HEAD:

- Global Invariant: intent preservation across the four reworded templates (nested-spawn warning, PASS/FAIL/BLOCKED contract)
- Skill-activation wording: template rewording correctness, and prompt-quality review
- Branch-race guidance: CODING.md Default content, do/SKILL.md cross-reference, and CLAUDE.md-adherence review
- Both ADRs: MADR template compliance, self-contained evidence, decision fidelity, and cross-linking
- Plugin version bump: patch-level increase confirmed against the pre-branch base
-


doodledood:
I'm trying to understand the problems this pr tries to solve help me out 

1. You had instances where /define encoded imprecise prompts in the criteria prompts that caused the verifier itself to not use some skill?

2. Not sure what this is about - verifiers shouldn't be doing anything destructive - only read only stuff or max create some temp artifacts. Do you think this is not clear enough from the define prompt skill?

Dev-iL:
**Point 1 — skill-activation wording**

- **Expected:** A templated `verify.prompt` like *"Activate the manifest-dev:review-code skill with dimension=code-bugs and review the change. PASS only if no LOW-or-higher findings."* was expected to make the verifying subagent literally invoke `review-code` — i.e., actually call the skill and let its rubric drive the verdict.
- **Actual:** Running that exact same prompt repeatedly against identical diffs, the verifier sometimes never invoked the skill at all — it read the diff itself with its own ad hoc commands and reported findings directly, skipping the skill's rubric entirely. This got *worse*, not better, once the prompt was embedded inside a full manifest rather than run standalone — which is the realistic way these prompts actually execute in production. In one captured instance this wasn't just a style difference: the run that skipped the skill returned a different verdict (FAIL) than every other run against the same underlying code (PASS, whether or not they'd invoked the skill) — so the ambiguity was concretely flipping outcomes, not just varying prose.
- **How the fix helps:** So yes — your reading in (1) is exactly right. The fix rewords the four templates `/define` copies into manifests from "Activate the X skill…" to an explicit anti-free-hand directive: *"Invoke the manifest-dev:review-code skill for real, with dimension=<dimension>; do not free-hand this review by reconstructing the rubric from memory instead of actually invoking the skill."* It keeps the same "invoke a skill" convention (rather than naming Claude Code's literal "Skill tool," which wouldn't translate to the other harnesses this ships to) but names the exact failure mode — free-handing, reconstructing from memory — so "activate" can't be satisfied by just reasoning about the skill instead of loading it.

**Point 2 — the worktree/branch-race guidance**

- This isn't about verifiers doing something destructive — you're right that they should stick to read-only inspection or their own scratch space, and nothing here changes that expectation. The gap is narrower: some Acceptance Criteria are legitimately designed to exercise a branch's actual checked-out files, not just diff them — e.g. "run the project's lint/pre-commit suite on branch X and confirm it passes." That kind of check needs the branch's files really materialized on disk; a `git diff` can't run a linter.
- **Expected:** That verifier runs `git checkout <branch>` in the working directory, runs its tooling, and reports the result — a normal, sanctioned way to do that kind of check.
- **Actual:** `/do` runs every criterion in the same phase concurrently, and by default *every* criterion is phase 1 unless the manifest author sets `phase:` explicitly — and all of them share one working directory (the orchestrator's `cwd`). So while that verifier is running `git checkout branch-A` (mutating HEAD and the working tree), sibling verifiers running at the same time were doing their own `git diff`/`git show` against what they assumed was a stable checkout — and their reads raced against this verifier's writes. That's a correctness bug in the *orchestration model*, not a case of a verifier overstepping its bounds.
- **How the fix helps:** it doesn't forbid this kind of check — it gives it a safe way to happen: isolate via a disposable `git worktree add <tmp-dir> <branch>` (a separate directory, so there's nothing shared to race on) instead of checking out in place, with per-criterion `phase:` separation as the fallback when a dedicated worktree isn't practical.
- **On your last question** — no, this genuinely wasn't documented anywhere before. I checked: neither `define`'s nor `do`'s skill files say anything about verifiers being read-only, or about the concurrency/shared-directory model at all. So it's not that the existing guidance was unclear — this axis (shared vs. isolated resources under concurrent execution) simply didn't have any guidance yet.

PR #219, opened 2026-07-05
Dev tools: behavior verification framework

Dev-iL:
## Summary

Adds an opt-in **empirical skill-behavior verification framework** (`behavior_lab`) under `manifest-dev-tools:prompt-engineering`, so a future skill/prompt wording change that claims to fix an observed behavior can be proven with real captured LLM traffic instead of only static review.

- `harness.py` — a minimal adapter interface (`configure_for_capture` / `invoke` / `decode_call`) with a working `ClaudeCodeHarness` that decodes real Anthropic SSE/chunked traffic into usage + text/tool_use/thinking blocks, plus `CodexHarness`/`PiHarness` stubs that raise `NotImplementedError` with the exact confirmed config extension points (`openai_base_url` in `~/.codex/config.toml`; `baseUrl` in `~/.pi/agent/models.json`).
- `proxy.py` — the generalized logging reverse-proxy used to capture real request/response traffic.
- `experiment.py` — generic `Arm` (harness + model/effort + prompt-construction config) and `Scenario` (caller-supplied task) abstractions, a runner that executes scenarios × arms through the harness adapter interface, plus session-json parsing: usage/cost rollup and session-to-scenario attribution over a captured run directory.
- `transcript.py` — turn-by-turn session reconstruction via the harness adapter's `decode_call`, session-scoped and run-scoped.
- `assertions.py` — `assert_tool_invoked(calls, tool_name, input_predicate=None)` and `diff_arms(results_by_arm, assertion)` for a one-call baseline-vs-amended compliance comparison.
- `references/EMPIRICAL_VERIFICATION.md` — the end-to-end workflow doc, linked from `prompt-engineering`'s `SKILL.md`.
- `manifest-dev-tools` plugin bumped `2.0.0` → `2.1.0`.

## What's in each commit

- `eb9af95` — the framework itself (all five modules, tests, reference doc, `SKILL.md` pointer, version bump).
- `850ec6f` — a fix for a real bug an adversarially-verified `change-intent` review caught in the first commit's own reference doc: the worked example paired `experiment.load_run_calls` (a usage-only summary with no `request_body`/`response_body`) with `Harness.decode_call` (which needs those bodies), silently producing empty `response_blocks` for every call — so `assert_tool_invoked` always returned `False` regardless of what the model actually did. Adds `transcript.load_run_entries`/`decode_calls_for_run` (the run-scoped counterpart to the existing session-scoped `decode_calls_for_session`), repoints the doc at it, and adds a regression test that pairs the fixed function directly with `assert_tool_invoked` across a multi-session run directory.

## Test plan

- [x] `python3 -m pytest claude-plugins/manifest-dev-tools/skills/prompt-engineering/scripts/behavior_lab/tests -q` — 27 passed (every module has ≥1 positive + 1 negative case; `CodexHarness`/`PiHarness` confirmed to raise `NotImplementedError`; fixtures are hand-written via `tmp_path`, no dependency on real captured data).
- [x] `ruff check` / `black --check` on `scripts/` — clean.
- [x] `mypy` — out of this repo's configured scope (`pyproject.toml`'s `[tool.mypy] files` only lists two `tests/` files); skipped per that scope, not silently ignored.
- [x] `git diff main...HEAD -- tests/cache-experiment/` — empty; nothing under that path changed.
- [x] `review-code` `context-file-adherence`, `change-intent` dimensions — both PASS (the latter after the fix above).
- [x] Manual: confirmed the doc's worked example's function calls match the real module signatures.

Manifest: `manifest-20260704-163918.md` (all 10 Acceptance Criteria / Global Invariants independently verified PASS).


doodledood:
First before I dig into it: I don't think it should be under the tools plugin - more like something new like manifest-dev-misc or -meta 

The -tools plugin is more for extra tools for dev work

Dev-iL:
Done - it's now under `manifest-dev-meta`

PR #221, opened 2026-07-08
Add a judgment layer (premise check) to PR review

doodledood:
## What

Adds a **judgment layer** to `review-pr` — a non-blocking, evidence-gated pass that asks whether a change earns its keep against the pain it solves, alongside the existing defect-finding review. It runs in both review modes and posts author-facing questions; it never blocks a merge.

## Why

The 13-dimension `review-code` fleet finds defects inside a change whose intent is taken as given — its narrow-lens reviewers never see the PR description/conversation, and its dimensions drop anything the author chose deliberately. So it structurally cannot ask *"should this exist, in this shape, at this footprint?"* The judgment layer adds exactly that, at the one surface that holds the stated pain (the PR description/conversation in no-manifest mode; the manifest's Intent in manifest mode).

## What it does

One question — *does this change, its approach, and its footprint earn their keep against the pain it solves, and is anything the pain requires missing?* — with seven evidence-gated triggers: necessity, pain-reconstructable, surface-proportionality, solution-shape, omission-vs-pain, irreversibility, precedent. Each fires only on concrete, nameable evidence, at whole-PR altitude (never line-level nitpicks); taste-level "I'd do it differently" stays dropped.

- **Own parallel subagent, consolidated by the holistic pass.** The judgment pass runs as its own wide-context subagent spawned in parallel with the fleet (fed the PR description/pain the narrow-lens fleet never sees), not folded into the holistic-coherence subagent. In no-manifest mode the holistic pass consolidates its findings alongside the fleet's; in manifest mode it is a strictly additive parallel step. It self-synthesizes one question per root before returning, so that holds in manifest mode too (which skips the holistic pass).
- **Distinct non-severity finding class.** Judgment findings sit off the defect severity scale (no low/med/high/critical) — a premise question is a different axis. They carry `{trigger, evidence, author-facing question}`, ride the existing voice/marker/posting path, and never block a merge, submit `request_changes`, auto-approve, or alter a manifest contract verdict.
- **Once per PR.** Fresh judgment *generation* is gated to once per PR, keyed on a `judgment` discriminator token in the hidden self-marker (`<!-- manifest-dev:review-pr judgment -->`) so a prior judgment finding is detectable — the base marker alone can't, since every review-pr comment carries it. Once a question has posted, later passes skip regeneration and never re-nag; thread advancement is never gated. A silent pass posts no marker, so on a sound PR it may re-run on later rounds — bounded and harmless (it stays silent). Accepted tradeoff: a premise concern introduced by a late commit after a question posted is not re-caught.
- **Premise-subsumption (no-manifest).** When a judgment finding questions whether a surface should exist, the holistic pass subordinates the fleet's moot-if-removed findings on that surface under the premise question rather than listing them as independent peers — without hard-deleting a severe independent defect. Not applicable in manifest mode (no fleet).
- **Both modes.** No-manifest mode consolidates through the holistic pass; manifest mode runs it as the independent premise check that catches what a define session missed, leaving contract verification untouched.

## Changes

- `review-pr/references/JUDGMENT_PASS.md` — new reference defining the layer: the question, the seven triggers with fires-on/never bounds, the evidence bar, the non-severity finding class, and synthesis.
- `review-pr/SKILL.md` + `references/MANIFEST_MODE.md` — parallel-subagent wiring in both modes, the once-per-PR gate and `judgment` marker discriminator, premise-subsumption in the holistic pass, plus cycle-summary reporting of posted questions.
- Two ADRs: `20260708-judgment-layer-is-a-review-time-premise-check` (premise-questioning is a review-time concern, not a define gate or a `/do` step — the manifest is the binding acceptance contract, and a non-blocking question has no gate to satisfy there) and `20260708-judgment-layer-runs-in-manifest-mode-as-a-premise-safety-net`.
- `CONTEXT.md` — "Judgment Layer" glossary term and relationship.
- Version bumps: `manifest-dev-tools` 2.0.0 → 2.1.0, Pi package 1.4.0 → 1.5.0; regenerated the Pi/Codex/OpenCode distributions.

## Verification

Built and verified through the manifest workflow: every acceptance criterion and global invariant was checked by an independent subagent (change-intent, review-prompt, code-design, plus reference/wiring/non-blocking/finding-class/once-per-PR/version/distribution/PR-mergeability checks). A first verification round caught real divergences in the once-per-PR mechanism — the base self-marker couldn't distinguish a judgment finding from any other review-pr comment, and manifest mode claimed a synthesis step it skipped — which were fixed (the `judgment` discriminator token and subagent self-synthesis) and re-verified green.


Dev-iL:
What about invoking "external advisors" (local models, different vendors) if available?

doodledood:
@Dev-iL sure since there are endless options though it shouldn't be a part of it but you can just run review-pr with a note telling it to run the judgement run with another model for the subagent or even a call to codex headless etc and it will just work

PR #257, opened 2026-08-09
Make the ADR methodology self-triggering, and add /init-context

doodledood:
The ADR index in this repo drifted five rows in three days — four records never added, one row still claiming a decision was current after a later one replaced it. Nothing caused it but silence: no document anywhere instructed maintaining the index. `ADR_FORMAT.md` had no index section and `WITH_DOCS.md`'s accept path stopped at writing the record. Every documented step was followed; the only one skipped was the one nobody wrote down.

So the methodology was not unclear, it was **under-triggered** — and a teammate not running manifest-dev had nothing to follow at all. This gives every step a trigger, ships the conventions as project knowledge, and adds a skill that installs the same wiring into any repository.

## Conventions become project knowledge

`ADR_FORMAT.md` is now self-sufficient: the bar for what deserves a record (folded in from `WITH_DOCS.md`) sits alongside the write-time mechanics, so a reader with no plugin can both decide and write. `docs/adr/CONVENTIONS.md` is this repo's copy, and **where a project has one, it governs** — the plugin defers on template, naming, lifecycle and index format. `WITH_DOCS.md` keeps only *cadence*, and loads the conventions at bootstrap, which is the first time anything in the workflow reads the ADR directory rather than only writing to it.

Also new there: an `Area` field, the index rebuild rules, the "one act, not three" write sequence, and the form a record takes when its rationale could not be recovered.

## The index is derived

It now holds nothing the records don't. `Area` backfilled across all 66 existing records; one relationship the index knew and its own record didn't written back; editorial link labels replaced with filenames so no column is un-derivable. Deleting `docs/adr/README.md` and rebuilding it from the records reproduces it byte for byte — that is the check, run by an independent verifier who wrote its own rebuilder from the spec.

No tooling ships. The value of a derived index is recoverability, not automation, and a markdown parser in every user's repo would buy determinism the specification already supplies.

Measured before deciding: the index costs ~45 tokens per row and ~3k in total, and a theme tree would not bound growth (19 of 62 rows sat in one Area). A split by live-versus-superseded is the move when size eventually justifies one.

## Triggers in the context file

`CLAUDE.md` gains the glossary residency rule (import, plus a session-start instruction for hosts without imports), when to open the corpus, and the three-step write act. The bare `docs/adr/` pointer is gone — a pointer with no trigger is exactly how the corpus went unread.

## /init-context

Installs those surfaces into any repository and seeds them from its history where there is any. It resolves the host's context file rather than assuming `CLAUDE.md`, ranks glossary candidates instead of sweeping and takes one batch approval before writing any, and marks records whose reasoning could not be recovered instead of inventing one. Mining is default-on with `--no-mine`; greenfield needs no flag because there is nothing to mine. It lives in the core plugin, since figure-out hands off to it and the plugins have no dependency edge.

Exercised against a clone of `kelseyhightower/envconfig` (132 commits, never used manifest-dev) and an empty repository. The brownfield run produced two records that deliberately differ in provenance — one with rationale recovered from a commit body, one carrying the not-recovered marker.

## Verification

23 gates, `per-gate` mode, one fresh independent verifier each. Three repair rounds: a skill that would have declared an existing ADR corpus non-conforming, an emission recipe that did not reproduce its own worked example, and a verbatim boundary that contradicted the instruction above it.

Four sub-threshold findings were handed over rather than repaired, and one pre-existing defect was found and deliberately left alone: the dist copies of `define/tasks/CODING.md` carry `CLAUDE.md` where the sync rules say those mentions should remap to `AGENTS.md`. It predates this branch and fixing it here would be work no criterion asked for.

Four ADRs record the decisions behind all of this.


Dev-iL:
Re "index is derived" - consider adding a python script to parse mdev-compatible ADRs deterministically.

PR #319, opened 2026-09-13
test(evals): behavioral eval suite for figure-out, with its noise floor measured

doodledood:
## What this adds

A 15-case behavioral eval suite for `figure-out` under `claude-plugins/manifest-dev/evals/`, plus
one skill fix it surfaced.

Every case runs twice — with the plugin and without it — so the number the suite reports is
**uplift (Δ)**, not pass rate. A case both arms pass measures nothing about the prompt and is
removed; one case was retired on exactly that ground.

```
claude plugin eval . --ablation with-without --judge-model sonnet --allow-tools Write Bash
```

## The one behavior change

`just-figure-out` wrote its investigation log to `figure-out-<UTC>.md`; `figure-out`'s `LOG.md`
specifies `figure-out-log-{timestamp}.md`. Running both skills left logs under two conventions in
one directory, and a glob for either missed the other. Both now write the same shape. This is the
only shipped-prompt change in the PR.

## The climb was attempted, measured, and reverted

The suite exists to hill-climb against, so it was used for that. One edit was made to the ask rule
in both skills — it said *the ask is the turn's open call* but never said which ask survives when
two compete, and turns were closing on two. It traced to `turn-discipline` failing the with-plugin
arm on two cases, `presses-one-crux` on a third, and `turn-discipline` on both lean mirrors.

**Verification could not tell whether it helped, and the over-fitting check came out backwards.**
The seven-case tuning mean went **+0.3016 → +0.2222**, failing the gate that asked for an increase.
The held-out set — the over-fitting detector — went **−0.2000 → +0.1000**. Over-fitting predicts
tuning *up* and held-out *down*; both moved the other way, so the split's own detector reports no
signal of any kind.

The *no-plugin arm* says why. It is the control: the plugin is not loaded, so no prompt edit can
reach it. Between the two measurements it moved by 0.30 or more on **five of fifteen cases** —
`05-move-on-evidence` by 0.667, `12j` by 0.500, `11` by 0.444, `04` and `12` by 0.333 each.

**An accident produced the cleanest number here.** A run whose wrapper script was killed had in
fact completed, so three cases were measured **twice against an identical prompt**, same command,
~40 minutes apart. Mean |Δ difference| **0.111**, max **0.167**, with one case's with-plugin score
moving 0.83 → 0.50 on text that did not change by a character. Comparing baseline to post-climb
confounds the edit with run-to-run variation; this isolates the variation.

Per-run score sd is **0.378** over 180 error-free runs, putting the standard error of the
seven-case tuning mean at σ·√(2/7n) = **0.117** at `runs: 3`. The move observed was −0.079 —
**0.68 of one standard error**.

So the edit was reverted and is not in this PR. A prompt change that cannot be shown to help does
not ship. I also did not re-run the tuning set hoping for a better mean: at that standard error
against a +0.3016 bar, roughly half of all re-runs of an *unchanged* skill would clear the gate by
chance, and passing it that way would be gaming the instrument.

**The complete data first argued the revert was wrong, and decomposition refuted that.** On every
subset except the one the gate measures, the edit looks better — held-out **+0.300 at 2.17 se**,
the only quantity in the dataset above two standard errors, on cases never tuned against. But Δ =
with − without, and only the with-arm can respond to a prompt edit:

| Subset | Δ move | = with-arm | − control-arm |
|---|---:|---:|---:|
| Tuning | −0.079 | −0.064 | +0.016 |
| Held-out | +0.300 | **+0.133** | **−0.167** |
| Whole suite | +0.122 | +0.074 | −0.048 |

More than half the held-out "improvement" is the **control getting worse**. On `05-move-on-evidence`
(Δ +0.667) and `12j` (Δ +0.500) — together +0.233 of the +0.300 — the with-plugin score did not
move by a single point; both are entirely the no-plugin arm collapsing from an unusually good
baseline. A Δ that improves because the control degraded is not an improvement, and it is the same
failure this suite was rebuilt to escape wearing different clothes. The rule is now written down
where the next climb will read it.

**This is the most useful thing the suite produced.** Resolving a change of that size needs
n = 2σ²/(7·se²) ≈ **26 runs per case per arm** — roughly nine times the cost per verification
(~$500 per full-suite run, against the $57 the three completion reports cost). Anyone using a Δ
from this suite to justify a prompt edit needs that number first, which is why `evals/README.md`
states it beside the baseline rather than in a footnote.

## What the baseline shows about figure-out

**The sign of Δ tracks what the scenario needs, not how hard it is.** Every case whose failure mode
is *under*-pressing comes out positive — a solution arriving pre-chosen (+0.89), no crux handed
over (+0.83), evidence that underdetermines (+0.56). Every case where the right answer is
*restraint* comes out negative — living with a flake whose fix costs three engineer-weeks (−0.50),
complying with a settled decision, accepting a diagnosis the evidence already supports (−0.17 each).

That is one mechanism, not two findings: the prompt biases toward more investigation, which is
exactly its value in the first column and exactly its cost in the second.

The sharpest instance is held out from tuning, so nothing in the climb could have produced or
concealed it — and **it is the one result the noise did not eat**. `12-living-with-it` came out at
**Δ −0.50 in both independent measurements**, with `prices-doing-nothing` failing the with-plugin
arm **3/3 each time** — six runs, six failures — while the no-plugin arm scored 0.667 and then
1.000. Both arms shifted upward between measurements and the gap did not close. `SKILL.md` says
plainly that living with the cost is a real option, *recommended as a full answer when it wins*.
**That is a real gap between what the prompt states and what it produces**, and it is the most
actionable thing here.

## The lean variant — evidence, not a recommendation

Four scenarios are mirrored between `figure-out` and `just-figure-out`, differing only in which
skill the prompt names, with byte-identical graders. Measured twice, **the ordering flipped**: the
full prompt led all three completed scenarios in the first measurement, the lean prompt led all
three in the second.

Neither column is a clean read — the lean skill changed twice between measurements, its baseline
was depressed by the log-path bug this PR fixes (a third of one case's score lost to a filename),
and both sit inside the noise above. **This suite cannot currently separate the two variants** on
three of the four scenarios.

The fourth is the exception, and it favours the lean prompt: on **living with it** the lean variant
beat the full one in both measurements (−0.33 vs −0.50, then +0.17 vs −0.50), and its
`prices-doing-nothing` passed once where the full prompt's never did across six runs. That is the
same restraint-side gap above, with less of it in the narrower prompt — one scenario, two
measurements, worth a targeted test rather than a conclusion.

The better lead is at grader level, because it is a claim about *which* rules survive compression
rather than an aggregate noise dominates: on the underdetermined scenario the lean prompt matched
the full prompt exactly — 3/3 on all three epistemic-discipline graders — while trailing on turn
shape, `turn-discipline` 1/3 against 3/3 on two scenarios. Testing that directly is far cheaper
than raising the run count on everything.

**This is evidence toward a graduation decision the owner holds.** It is not a recommendation to
graduate or retire either variant, and four scenarios — one blocked, one bug-affected — could not
support one.

## Caveats a reviewer should have

- **The baseline is already stale in one respect.** While the suite was running, `main` rewrote the
  `figure-out` paragraph that states the ask rule — the same paragraph the climb targeted — and
  added a momentum rule to `just-figure-out`. Every number here describes the prompts as they stood
  before those commits. The suite is durable; the baseline is not, and it needs re-measuring before
  the next climb.
- **`figure-out` does not reliably auto-fire.** Across 9 phrasings and 12 runs, only an explicit
  "use the figure-out skill for this" fired it 6/6. The prompts say so rather than working around
  it. Prefixing with the slash command was tried and disqualified: it makes the no-plugin arm
  return zero turns, manufacturing a false +1.0 Δ.
- **One gate fails: AC-3.2**, the tuning set improving. Every other criterion passes, including the
  over-fitting check. The evidence says that gate asks the instrument for a resolution it does not
  have, rather than that the edit was bad — but it is reported as failed, not rewritten.
- **Three graders were recalibrated, every fix loosening them.** Two shared one defect: a
  requirement followed by prose exemptions, which a sonnet judge applied while skipping the
  exemptions. Both were rewritten to enumerate and classify before judging. That cost a second
  baseline, which is the freeze rule working as intended.
- **Eight scored graders pass in both arms on every baseline run** and are declared as floor
  checks rather than counted as uplift. A grader passing in both arms says the base model already
  carries that discipline unprompted — information, not a defect — but it moves no Δ. They stay
  because they are what any "press harder" tuning breaks first: `04` and `05` are a mirror pair no
  constant policy passes, and `07`'s is the over-trigger floor.
- **Run evidence is committed as JSON**, so every number in `evals/README.md` is recomputable from
  the repository — per-grader verdicts, scores, costs and turn counts for both arms of every run.
  The rendered HTML reports and traces are not: the HTML duplicates the JSON at twice the size,
  and traces hold full model output.

## Verification

`ruff` clean · `black` 11 files unchanged · `mypy` no issues · `pytest` 109 passed, 12 skipped.
`dist/codex` synced via `scripts/sync_dist.py`; plugin 11.9.0 → 11.9.1 across the Claude, Codex and
OpenCode manifests, Pi package 8.10.0 → 8.10.1.

https://claude.ai/code/session_01Q29FHsy8s5gUZNa3X7zfai


Dev-iL:
Shana tova! Didn't you want to put similar things under `manifest-dev-meta` (#219)?

doodledood:
Closing: the eval scenarios in this branch were drawn from real project material. Replacing with fictional equivalents and reopening a fresh PR.

